PR #1, opened 2020-08-25
Added optional DB port parameter for MySQL store.

yuriio147:


stargazer/plugins/store/mysql/mysql_store.cpp
```diff
@@ -146,13 +147,25 @@ if (ParseParam(s.moduleParams, "server", dbHost) < 0 &&
     ParseParam(s.moduleParams, "dbhost", dbHost) < 0)
     return -1;
 
+// not required
+std::string dbPortAsString;
+if (ParseParam(s.moduleParams, "port", dbPortAsString) == 0 ||
+    ParseParam(s.moduleParams, "dbport", dbPortAsString) == 0)
```
madf, line 153: Why two different names?

madf: Okay, I see, you just follow the existing naming scheme.

PR #6, opened 2024-12-02
Stargazer

HelenMamontova:
Plugin Radius added.
Parsing command line parameters added.
Implemented starting Stargazer without starting daemon.


projects/stargazer/main.cpp
```diff
@@ -208,6 +209,21 @@ void KillExecuters()
         ++pid;
     }
 }
+
+void PrintHelp(const std::string& programName)
+{
+    std::cout << "Usage: " << programName << "[-h/--help] [-v/--version] [-f/--foreground] [../dist]\n"
```
madf: `[../dist]` -> `<conf-dir-path>`

projects/stargazer/main.cpp
```diff
@@ -208,6 +209,21 @@ void KillExecuters()
...
+              << "\t --help, -h            - print this help;\n"
+              << "\t --version, -v         - print version;\n"
+              << "\t --foreground, -f      - do not go into background;\n"
+              << "\t ../dist               - path to the directory where the configuration file is located .\n";
```
madf: See above

projects/stargazer/main.cpp
```diff
@@ -223,7 +239,33 @@ int main(int argc, char* argv[])
...
+        {
+            PrintHelp(argv[0]);
+            return 0;
+        }
+        if (arg == "--version" || arg == "-v")
+        {
+            PrintVersion(argv[0]);
+            return 0;
+        }
+        if (arg == "../dist")
```
madf: This is plain wrong

projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -0,0 +1,19 @@
...
+
+extern "C" STG::Plugin* GetPlugin()
+{
+    static RADIUS plugin;
+    return &plugin;
+}
+
+std::string RADIUS::GetVersion() const
+{
+    return "Radius authorizator v.1.0";
```
madf: Please remove 'authorizator'. This is not an authorization plugin, although it can optionally perform authorization. Just 'Radius' is fine.

madf (changes requested):

projects/stargazer/main.cpp
```diff
@@ -208,6 +209,21 @@ void KillExecuters()
...
     }
 }
+
+void PrintHelp(const std::string& programName)
+{
+    std::cout << "Usage: " << programName << "[-h/--help] [-v/--version] [-f/--foreground] [<conf-dir-path>]\n"
+              << "\t --help, -h            - print this help;\n"
+              << "\t --version, -v         - print version;\n"
+              << "\t --foreground, -f      - do not go into background;\n"
+              << "\t <conf-dir-path>       - path to the directory where the configuration file is located .\n";
```
madf: Extra whitespace before '.'.

projects/stargazer/main.cpp
```diff
@@ -223,7 +239,39 @@ int main(int argc, char* argv[])
...
+            if (arg == "--version" || arg == "-v")
+            {
+                PrintVersion(argv[0]);
+                return 0;
+            }
+            if (arg == "--foreground" || arg == "-f")
+            {
+                noDaemon = true;
+
+                if (argc == i + 2)
```
madf: Why do you need this condition? Wouldn't the following command line argument go into the 'else' branch on the next cycle iteration?

madf (changes requested):

madf (changes requested):
I am okay with your changes. However, there is a macro out there: NO_DAEMON. It used to control the way we start stargazer (whether to make it a daemon or not). Now, with your changes, we control it in runtime by command line arguments. We need to adjust the logic driven by the macro to use command line argument, and completely eliminate the macro.

projects/stargazer/main.cpp
```diff
@@ -367,9 +356,7 @@ int main(int argc, char* argv[])
     WriteServLog("Stg started successfully.");
     WriteServLog("+++++++++++++++++++++++++++++++++++++++++++++");
 
-#ifndef NO_DAEMON
     creat(startFile.c_str(), S_IRUSR);
```
madf: Do we still need to create the file in no-daemon mode?

projects/stargazer/main.cpp
```diff
@@ -279,9 +270,7 @@ int main(int argc, char* argv[])
         return -1;
     }
 
-#ifndef NO_DAEMON
     const auto startFile = settings.GetConfDir() + START_FILE;
```
madf: It seems that `startFile` is only used in `creat`. It makes sense to move it closer to the point of use.

madf (changes requested):

projects/stargazer/main.cpp
```diff
@@ -356,7 +354,10 @@ int main(int argc, char* argv[])
     WriteServLog("Stg started successfully.");
     WriteServLog("+++++++++++++++++++++++++++++++++++++++++++++");
 
-    creat(startFile.c_str(), S_IRUSR);
+    const auto startFile = settings.GetConfDir() + START_FILE;
```
madf: And if `noDaemon` is `true`, we do not need this variable either.

projects/stargazer/main.cpp
```diff
@@ -223,7 +230,34 @@ int main(int argc, char* argv[])
         return 1;
     }
 
-    SettingsImpl settings(argc == 2 ? argv[1] : "");
+    std::string path;
+    bool noDaemon(false);
```
madf, line 234: Consider `bool noDaemon = false;`, it looks more naturally.

madf (changes requested):

PR #7, opened 2025-01-03
Radius

HelenMamontova:
Implemented so that the mod_radius works as a server in radius.

### DIFF
--- a/projects/stargazer/plugins/other/radius/radius.cpp
+++ b/projects/stargazer/plugins/other/radius/radius.cpp
@@ -45,9 +45,9 @@ int RADIUS::Run(std::stop_token token)
 
     try
     {
-        boost::asio::io_service io_service;
-        Server server(io_service, "secret", 1812, "/usr/share/freeradius/dictionary");
-        io_service.run();
+        boost::asio::io_service ioService;
+        Server server(ioService, "secret", 1812, "/usr/share/freeradius/dictionary");
+        ioService.run();
     }
     catch (const std::exception& e)
     {
--- a/projects/stargazer/plugins/other/radius/radius.cpp
+++ b/projects/stargazer/plugins/other/radius/radius.cpp
@@ -31,19 +31,10 @@ int RADIUS::Start()
 
 int RADIUS::Stop()
 {
-    std::lock_guard lock(m_mutex);
-
-    if (!m_thread.joinable())
-        return 0;
-
-    m_thread.request_stop();
-
-
-    if (isRunning)
-        m_thread.detach();
-    else
+    if (m_thread.joinable())
         m_thread.join();
 
+    m_thread.request_stop();
     return 0;
 }
 
--- a/projects/stargazer/plugins/other/radius/radius.cpp
+++ b/projects/stargazer/plugins/other/radius/radius.cpp
@@ -62,8 +62,7 @@ int RADIUS::Run(std::stop_token token)
     }
     catch (const std::exception& e)
     {
-        std::string except(e.what());
-        m_errorStr = "Exception in RADIUS::Run(): " + except;
+        m_errorStr = "Exception in RADIUS::Run(): " + std::string(e.what());
         m_logger("Exception in RADIUS:: Run(): %s", e.what());
         printfd(__FILE__, "Exception in RADIUS:: Run(). Message: '%s'\n", e.what());
     }
--- a/projects/stargazer/plugins/other/radius/radius.cpp
+++ b/projects/stargazer/plugins/other/radius/radius.cpp
@@ -35,10 +35,12 @@ int RADIUS::Start()
 
 int RADIUS::Stop()
 {
-    if (m_thread.joinable())
-        m_thread.join();
+    if (!m_thread.joinable())
+        return 0;
 
     m_thread.request_stop();
+
+    m_thread. join();
     return 0;
 }
 
--- a/projects/stargazer/plugins/other/radius/radius.cpp
+++ b/projects/stargazer/plugins/other/radius/radius.cpp
@@ -43,19 +43,17 @@ int RADIUS::Run(std::stop_token token)
     std::lock_guard lock(m_mutex);
     isRunning = true;
 
-    while (!token.stop_requested())
+    try
     {
-        try
-        {
-            boost::asio::io_service io_service;
-            Server server(io_service, "secret", 1812, "/usr/share/freeradius/dictionary");
-            io_service.run();
-        }
-        catch (const std::exception& e)
-        {
-            std::cerr << "Exception: " << e.what() <<"\n";
-        }
+        boost::asio::io_service io_service;
+        Server server(io_service, "secret", 1812, "/usr/share/freeradius/dictionary");
+        io_service.run();
     }
+    catch (const std::exception& e)
+    {
+        std::cerr << "Exception: " << e.what() <<"\n";
+    }
+
     isRunning = false;
     return 0;
 }
--- a/projects/stargazer/plugins/other/radius/radius.cpp
+++ b/projects/stargazer/plugins/other/radius/radius.cpp
@@ -25,32 +25,46 @@ RADIUS::RADIUS()
 
 int RADIUS::Start()
 {
-    isRunning = true;
-    m_thread = std::jthread([this](){ Run(); });
+    m_thread = std::jthread([this](auto token){ Run(std::move(token)); });
     return 0;
 }
 
 int RADIUS::Stop()
 {
     std::lock_guard lock(m_mutex);
-    isRunning = false;
+
+    if (!m_thread.joinable())
+        return 0;
+
+    m_thread.request_stop();
+
+
+    if (isRunning)
+        m_thread.detach();
+    else
+        m_thread.join();
+
     return 0;
 }
 
-int RADIUS::Run()
+int RADIUS::Run(std::stop_token token)
 {
-    std::string secret("secret");
-    uint16_t port = 1812;
+    std::lock_guard lock(m_mutex);
+    isRunning = true;
 
-    try
+    while (!token.stop_requested())
     {
-        boost::asio::io_service io_service;
-        Server server(io_service, secret, port, "/usr/share/freeradius/dictionary");
-        io_service.run();
+        try
+        {
+            boost::asio::io_service io_service;
+            Server server(io_service, "secret", 1812, "/usr/share/freeradius/dictionary");
+            io_service.run();
+        }
+        catch (const std::exception& e)
+        {
+            std::cerr << "Exception: " << e.what() <<"\n";
+        }
     }
-    catch (const std::exception& e)
-    {
-        std::cerr << "Exception: " << e.what() <<"\n";
-    }
+    isRunning = false;
     return 0;
 }
--- a/projects/stargazer/plugins/other/radius/radius.cpp
+++ b/projects/stargazer/plugins/other/radius/radius.cpp
@@ -37,9 +37,9 @@ int RADIUS::Stop()
     return 0;
 }
 
-int Run()
+int RADIUS::Run()
 {
-    std::string secret;
+    std::string secret("secret");
     uint16_t port = 1812;
 
     try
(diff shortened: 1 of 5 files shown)


projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -17,3 +23,34 @@ RADIUS::RADIUS()
...
+{
+    isRunning = true;
+    m_thread = std::jthread([this](){ Run(); });
+    return 0;
+}
+
+int RADIUS::Stop()
+{
+    std::lock_guard lock(m_mutex);
+    isRunning = false;
```
madf: `isRunning` must be set to `false` in the `Run` method, before exiting.

Also, see the `ping` plugin and look how `stop_token` and `m_thread.request_stop()` are used. In particular, notice that we join the thread in `Stop` after requesting stop (or detaching it in case we failed to stop the thread function).

projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -17,3 +23,34 @@ RADIUS::RADIUS()
...
+    return 0;
+}
+
+int RADIUS::Run()
+{
+    std::string secret("secret");
+    uint16_t port = 1812;
```
madf: Pass those values directly.

madf (changes requested):

projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -17,3 +23,48 @@ RADIUS::RADIUS()
...
+int RADIUS::Stop()
+{
+    std::lock_guard lock(m_mutex);
+
+    if (!m_thread.joinable())
+        return 0;
+
+    m_thread.request_stop();
+
+
```
madf, line 47: Extra empty line.

projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -17,3 +23,48 @@ RADIUS::RADIUS()
...
 
+int RADIUS::Start()
+{
+    m_thread = std::jthread([this](auto token){ Run(std::move(token)); });
+    return 0;
+}
+
+int RADIUS::Stop()
+{
+    std::lock_guard lock(m_mutex);
```
madf: You do not need to hold the mutex here.

projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -17,3 +23,48 @@ RADIUS::RADIUS()
...
+
+    if (!m_thread.joinable())
+        return 0;
+
+    m_thread.request_stop();
+
+
+    if (isRunning)
+        m_thread.detach();
```
madf: Let's just go for `join` for now. Unconditionally.

projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -17,3 +23,48 @@ RADIUS::RADIUS()
...
+    else
+        m_thread.join();
+
+    return 0;
+}
+
+int RADIUS::Run(std::stop_token token)
+{
+    std::lock_guard lock(m_mutex);
```
madf: This lock has a function-level scope. It will be released when the function finishes. It means that no one can access `isRunning` while the function is running. You should hold locks only when you read or write a variable, `isRunning` in particular, and release them as soon as possible - after reading or writing.

projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -17,3 +23,48 @@ RADIUS::RADIUS()
...
+    isRunning = true;
+
+    while (!token.stop_requested())
```
madf: This will not work. `io_service.run()` will run indefinitely and you won't reach the second iteration of this cycle. Let's just skip checking the `token` fro now.

projects/stargazer/plugins/other/radius/radius.h
```diff
@@ -24,7 +26,10 @@ namespace STG
 
         private:
             mutable std::string errorStr;
+            std::jthread m_thread;
+            std::mutex m_mutex;
             bool isRunning;
```
madf: I know that the old plugins use `isRunning` and `errorStr`, but in the new code let's stick to the `m_` naming scheme for the private members. So please change them to `m_running` and `m_errorStr`.

projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -17,3 +23,48 @@ RADIUS::RADIUS()
...
+    {
+        try
+        {
+            boost::asio::io_service io_service;
```
madf: Let's call it `ioService` or `ios`. We do not use underscore in variable and function names.

projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -17,3 +23,48 @@ RADIUS::RADIUS()
...
+            Server server(io_service, "secret", 1812, "/usr/share/freeradius/dictionary");
+            io_service.run();
+        }
+        catch (const std::exception& e)
+        {
+            std::cerr << "Exception: " << e.what() <<"\n";
```
madf: Avoid using `std::cout` and `std::cerr`. See how the other plugins report errors.

projects/stargazer/plugins/other/radius/server.h
```diff
@@ -0,0 +1,23 @@
+#pragma once
+
+#include "radproto/socket.h"
+#include "radproto/packet.h"
+#include "radproto/dictionaries.h"
+#include <boost/asio.hpp>
+#include <optional>
+#include <cstdint> //uint8_t, uint32_t
+
+class Server
```
madf: Please put it into the `STG` namespace.

madf (changes requested):

projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -14,6 +23,48 @@ std::string RADIUS::GetVersion() const
...
+
+    try
+    {
+        boost::asio::io_service ioService;
+        Server server(ioService, "secret", 1812, "/usr/share/freeradius/dictionary");
+        ioService.run();
+    }
+    catch (const std::exception& e)
+    {
+        m_errorStr = "Exceptoin by Run()";
```
madf: 'Exception' -> 'Exception' (misprint).

'by Run()' -> 'in RADIUS::Run'

Also please add `e.what()`.

Same below.

projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -14,6 +23,48 @@ std::string RADIUS::GetVersion() const
...
+int RADIUS::Start()
+{
+    m_thread = std::jthread([this](auto token){ Run(std::move(token)); });
+    return 0;
+}
+
+int RADIUS::Stop()
 {
+    if (m_thread.joinable())
+        m_thread.join();
```
madf: You must join a thread only after you have requested it to stop. Because before requesting, the join would hang forever, because the thread does not know it is expected to stop. And because the join hangs, you can not request the thread to stop.

The logic here should be as follows:
 - if the thread is not joinable, nothing to be done here, just return;
 - otherwise, request stop;
 - join the thread to wait until it stops.

projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -14,6 +23,48 @@ std::string RADIUS::GetVersion() const
...
+
+    m_thread.request_stop();
+    return 0;
+}
+
+void RADIUS::SetRunning(bool val)
+{
+    std::lock_guard lock(m_mutex);
```
madf: All `std::lock_guard` instances can and should be `const`.

projects/stargazer/plugins/other/radius/radius.h
```diff
@@ -10,21 +14,26 @@ namespace STG
...
             uint16_t GetStopPosition() const override { return 0; }
 
             int SendMessage(const Message & msg, uint32_t ip) const override { return 0; }
 
         private:
-            mutable std::string errorStr;
-            bool isRunning;
+            mutable std::string m_errorStr;
+            std::jthread m_thread;
+            std::mutex m_mutex;
```
madf: A little pro-tip. Always put the mutex first in the member list. This way you ensure that it is initialized when you use it.

I had an issue in the past when I put the mutex after the thread, like you here. And I started the thread right inside the initialization, and it ran before the mutex was initialized, and inside the thread function I tried to lock the mutex and everything crashed, because the mutex was not initialized. It was a really tricky issue to find, because most time the thread starts with a little delay, and the mutex has a chance to be initialized.

projects/stargazer/plugins/other/radius/radius.h
```diff
@@ -10,21 +14,26 @@ namespace STG
...
+            bool m_running;
 
+            int Run(std::stop_token token);
+            PluginLogger m_logger;
```
madf: Please put the `m_logger` after the `m_running`, in the member list.

madf (changes requested):

projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -60,9 +62,10 @@ int RADIUS::Run(std::stop_token token)
     }
     catch (const std::exception& e)
     {
-        m_errorStr = "Exceptoin by Run()";
-        m_logger("Exception by Run: %s", e.what());
-        printfd(__FILE__, "Exception by Run. Message: '%s'\n", e.what());
+        std::string except(e.what());
+        m_errorStr = "Exception in RADIUS::Run(): " + except;
```
madf: Consider `+ std::string(e.what());`

madf (changes requested):

projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -14,6 +23,50 @@ std::string RADIUS::GetVersion() const
...
 }
 
+int RADIUS::Stop()
+{
+    if (!m_thread.joinable())
+        return 0;
+
+    m_thread.request_stop();
+
+    m_thread. join();
```
madf: A whitespace after '.'

projects/stargazer/plugins/other/radius/radius.h
```diff
@@ -10,21 +14,26 @@ namespace STG
         public:
             RADIUS();
 
-            int Start() override { return 0; }
-            int Stop() override { return 0; }
+            int Start() override;
+            int Stop() override;
             int Reload(const ModuleSettings & /*ms*/) override { return 0; }
-            bool IsRunning() override { return isRunning; }
+            bool IsRunning() override { return m_running; }
```
madf: Returning `m_running` must be guarded by mutex.

madf (changes requested):

projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -50,6 +50,12 @@ void RADIUS::SetRunning(bool val)
     m_running = val;
 }
 
+bool RADIUS::IsRunning()
+    {
```
madf: Bad indentation.

madf (changes requested):

projects/stargazer/plugins/other/radius/server.h
```diff
@@ -0,0 +1,26 @@
+#pragma once
+
+#include "radproto/socket.h"
+#include "radproto/packet.h"
+#include "radproto/dictionaries.h"
+#include <boost/asio.hpp>
+#include <optional>
+#include <cstdint> //uint8_t, uint32_t
+                   //
```
madf: Unused `//`

madf (changes requested):

projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -14,6 +23,56 @@ std::string RADIUS::GetVersion() const
 }
 
 RADIUS::RADIUS()
+    : m_logger(PluginLogger::get("radius"))
```
madf: `m_running` must be also initialized here.

madf (changes requested):

PR #9, opened 2025-02-28
Radius

HelenMamontova:
Stop stargazer by ctrl+c implemented.


projects/stargazer/plugins/other/radius/radius.h
```diff
@@ -72,5 +76,7 @@ namespace STG
             std::jthread m_thread;
 
             PluginLogger m_logger;
+
+            std::unique_ptr<Server> m_server{};
```
madf: You do not need an initializer here.

projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -113,6 +88,7 @@ int RADIUS::Stop()
     if (!m_thread.joinable())
         return 0;
 
+    m_server->stop();
```
madf: This should go after the `m_thread.request_stop()`. I try to explain why.

First, you do not use `stop_token`. You take it in lambda when you create a thread, and you pass it to the `Run`. And it is not used there in any way. It means that `m_thread.request_stop()` has no effect now. Your token turns into a 'stop requested' state, but you don't check it.

Second, when you ask the `Server` to stop, you simply close the socket. And the receiving operation that is currently waiting for data is ended with an error ('operation aborted' or something). This is not terrible, yet it is not great. When a user stops the server, the last thing he expects is to receive an obscure 'operation aborted' error in the log.

Third, your `Server` is polluted with `std::cout`. Please don't use it here. For debug purposes use `printfd` and for user interaction use the logger (`m_logger` in your code).

Taking all that into account, here is what I think. Communication errors can happen, and we should report them. Unless the errors are made by us to interrupt receiving data. When the `Server::handleReceive` is called, we need to see if we are stopping first. Because if we are, the errors do not matter and we can simply return. And if we are not, we should check for errors and report them, if there are any. The only way to see if we are stopping is to check for the stop token. There is no stop token in the `Server`, but imagine that you have it there. And you check it. And if you call the `Server` to stop before requesting the thread to stop, the stop token would not report that we are stopping.

To sum the thing up:
 - you need to reorder `m_server->stop()` and `m_thread.request_stop()`;
 - you need to check that `m_server` contains a value before calling `stop()`;
 - you need to have a stop token in the `Server`;
 - you need to check the token in handlers before making any further steps;
 - you need to replace all `std::cout` in the `Server` with either `printfd` or `m_logger` or both.

projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -130,3 +106,24 @@ void RADIUS::SetRunning(bool val)
...
 }
+
+int RADIUS::Run(std::stop_token token)
+{
+    SetRunning(true);
+
+    try
+    {
+        if (!m_server)
+           m_server = std::make_unique<Server> (m_ioService, m_radSettings.GetSecret(), m_radSettings.GetPort(), m_radSettings.GetDictionaries());
```
madf: Extra whitespace after the `>` and before the`(`.

madf (changes requested):

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -1,14 +1,16 @@
...
 using STG::Server;
 using boost::system::error_code;
 
-Server::Server(boost::asio::io_service& io_service, const std::string& secret, uint16_t port, const std::string& filePath)
+Server::Server(boost::asio::io_service& io_service, const std::string& secret, uint16_t port, const std::string& filePath, std::stop_token token)
     : m_radius(io_service, secret, port),
-      m_dictionaries(filePath)
+      m_dictionaries(filePath),
+      m_token(token),
+      m_logger(PluginLogger::get("radius"))
```
madf: It's better to take it from the `RADIUS`.

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -54,22 +56,28 @@ RadProto::Packet Server::makeResponse(const RadProto::Packet& request)
...
 {
     if (ec)
-        std::cout << "Error asyncSend: " << ec.message() << "\n";
+    {
+        m_logger("Error asyncSend: %s", ec.message());
+        printfd(__FILE__, "Error asyncSend: '%s'\n", ec.message());
+    }
 
-    startReceive();
+    if (!m_token.stop_requested())
```
madf: I think you should check the token before checking the error code. When you close the socket and there is a pending send or receive operation, it will result in error. However, when we are stopping, this error is expected and is not something we want to report in the log.

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -54,22 +56,28 @@ RadProto::Packet Server::makeResponse(const RadProto::Packet& request)
...
+        startReceive();
 }
 
 void Server::handleReceive(const error_code& error, const std::optional<RadProto::Packet>& packet, const boost::asio::ip::udp::endpoint& source)
 {
     if (error)
```
madf, line 75: You should check for stop token here as well.

projects/stargazer/plugins/other/radius/server.h
```diff
@@ -3,7 +3,10 @@
 #include "radproto/socket.h"
 #include "radproto/packet.h"
 #include "radproto/dictionaries.h"
+#include "stg/logger.h"
 #include <boost/asio.hpp>
+#include <stop_token>
+#include <jthread.hpp>
```
madf: Why do you need `jthread.hpp` here?

projects/stargazer/plugins/other/radius/server.h
```diff
@@ -3,7 +3,10 @@
 #include "radproto/socket.h"
 #include "radproto/packet.h"
 #include "radproto/dictionaries.h"
+#include "stg/logger.h"
 #include <boost/asio.hpp>
+#include <stop_token>
```
madf: It should be `stop_token.hpp`. We use a 3rd-party library for `std::jthread` and `std::stop_token` because compilers may not have them yet.

madf (changes requested):

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -10,7 +10,7 @@ Server::Server(boost::asio::io_service& io_service, const std::string& secret, u
     : m_radius(io_service, secret, port),
       m_dictionaries(filePath),
       m_token(token),
-      m_logger(PluginLogger::get("radius"))
+      m_logger(PluginLogger::get("RADIUS"))
```
madf: As I said before, you should use logger from the `RADIUS`, not get a new one from `PluginLogger::get`. Pass it from the parent by ref and store by ref.

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -55,33 +55,36 @@ RadProto::Packet Server::makeResponse(const RadProto::Packet& request)
 
 void Server::handleSend(const error_code& ec)
 {
+    if (!m_token.stop_requested())
+        startReceive();
```
madf: No.

1. Check if stop is requested. If it is - just return.
2. Check error code.
3. Start receive.

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -55,33 +55,36 @@ RadProto::Packet Server::makeResponse(const RadProto::Packet& request)
...
     }
-
-    if (!m_token.stop_requested())
-        startReceive();
 }
 
 void Server::handleReceive(const error_code& error, const std::optional<RadProto::Packet>& packet, const boost::asio::ip::udp::endpoint& source)
 {
-    if (error)
+    if (!m_token.stop_requested())
```
madf: Same as in `handleSend`.

madf (changes requested):

projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -130,3 +108,24 @@ void RADIUS::SetRunning(bool val)
...
 }
+
+int RADIUS::Run(std::stop_token token)
+{
+    SetRunning(true);
+
+    try
+    {
+        if (!m_server)
+           m_server = std::make_unique<Server> (m_ioService, m_radSettings.GetSecret(), m_radSettings.GetPort(), m_radSettings.GetDictionaries(), token, m_logger);
```
madf: Extra whitespace between '>' and '('.

projects/stargazer/plugins/other/radius/server.cpp
madf: What about this?

projects/stargazer/plugins/other/radius/server.cpp
madf: And what about this?

madf (changes requested):

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -55,36 +55,35 @@ RadProto::Packet Server::makeResponse(const RadProto::Packet& request)
...
     {
-        if (error)
-        {
-            m_logger("Error asyncReceive: %s", error.message());
-            printfd(__FILE__, "Error asyncReceive: '%s'\n", error.message());
-            return;
-        }
+        m_logger("Error asyncReceive: %s", error.message());
+        printfd(__FILE__, "Error asyncReceive: '%s'\n", error.message());
+        return;
```
madf: It's interesting. In `handleSend` on error you log it and proceed. In `handleReceive` on error you log it and stop the server. Error handling should be uniform, you either proceed or stop. And since this is a server and it should be robust, it is better to proceed. Please make changes accordingly.

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -55,36 +55,35 @@ RadProto::Packet Server::makeResponse(const RadProto::Packet& request)
...
-        {
-            m_radius.asyncSend(makeResponse(*packet), source, [this](const auto& ec){ handleSend(ec); });
-        }
+    if (packet == std::nullopt)
+    {
+        m_logger("Error asyncReceive: the request packet is missing\n");
+        printfd(__FILE__, "Error asyncReceive: the request packet is missing\n");
+        return;
     }
+    else
```
madf: You do not need `else` if the previous branch has a `return`.

projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -130,3 +108,24 @@ void RADIUS::SetRunning(bool val)
...
 }
+
+int RADIUS::Run(std::stop_token token)
+{
+    SetRunning(true);
+
+    try
+    {
+        if (!m_server)
+           m_server = std::make_unique<Server>(m_ioService, m_radSettings.GetSecret(), m_radSettings.GetPort(), m_radSettings.GetDictionaries(), token, m_logger);
```
madf: `std::move(token)`

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -1,18 +1,31 @@
...
 
 using STG::Server;
 using boost::system::error_code;
 
-Server::Server(boost::asio::io_service& io_service, const std::string& secret, uint16_t port, const std::string& filePath)
+Server::Server(boost::asio::io_service& io_service, const std::string& secret, uint16_t port, const std::string& filePath, std::stop_token token, PluginLogger& logger)
     : m_radius(io_service, secret, port),
-      m_dictionaries(filePath)
+      m_dictionaries(filePath),
+      m_token(token),
```
madf: `std::move(token)`

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -42,27 +55,33 @@ RadProto::Packet Server::makeResponse(const RadProto::Packet& request)
...
 {
-    if (ec)
-        std::cout << "Error asyncSend: " << ec.message() << "\n";
+    if (m_token.stop_requested())
+        return;
 
+    if (ec)
+    {
+        m_logger("Error asyncSend: %s", ec.message());
+        printfd(__FILE__, "Error asyncSend: '%s'\n", ec.message());
```
madf: `ec.message()` returns `std::string`, while '%s' requires `const char*`. Please add `.c_str()` here and in other places.

madf (changes requested):

PR #11, opened 2025-04-01
Radius.

HelenMamontova:
Function findUser added in the class Server of Radius plugin.


projects/stargazer/plugins/other/radius/radius.h
```diff
@@ -18,6 +18,12 @@ namespace STG
 {
     struct Settings;
 
+    class Users;
+
+    using UserPtr = STG::User*;
+    using ConstUserPtr = const User*;
```
madf: `UserPtr` and `ConstUserPtr` are not used.

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -48,6 +51,13 @@ RadProto::Packet Server::makeResponse(const RadProto::Packet& request)
...
     vendorSpecific.push_back(RadProto::VendorSpecific(m_dictionaries.vendorCode("Dlink"), m_dictionaries.vendorAttributeCode("Dlink", "Dlink-User-Level"), vendorValue));
 
+    if (findUser(request))
+    {
+        m_logger("Error findUser\n");
+        printfd(__FILE__, "Error findUser\n");
+        return RadProto::Packet(RadProto::ACCESS_REJECT, request.id(), request.auth(), attributes, vendorSpecific);
+    }
+
     if (request.type() == RadProto::ACCESS_REQUEST)
```
madf: You should check request type before looking for a user.

projects/stargazer/plugins/other/radius/server.h
```diff
@@ -11,10 +14,16 @@
 
 namespace STG
 {
+    class Users;
+
+    using UserPtr = STG::User*;
+    using ConstUserPtr = const User*;
```
madf: `UserPtr` and `ConstUserPtr` are not used.

projects/stargazer/plugins/other/radius/server.h
```diff
@@ -3,6 +3,9 @@
 #include "radproto/socket.h"
 #include "radproto/packet.h"
 #include "radproto/dictionaries.h"
+#include "stg/users.h"
+#include "stg/user.h"
+#include "stg/user_property.h"
```
madf: `user.h` and `user_property.h` are not used in this file.

projects/stargazer/plugins/other/radius/server.h
```diff
@@ -3,6 +3,9 @@
 #include "radproto/socket.h"
 #include "radproto/packet.h"
 #include "radproto/dictionaries.h"
+#include "stg/users.h"
```
madf: `users.h` is not used in this file, you forward-declare `Users` below.

projects/stargazer/plugins/other/radius/server.h
```diff
@@ -11,10 +14,16 @@
...
+
+    using UserPtr = STG::User*;
+    using ConstUserPtr = const User*;
+
     class Server
     {
         public:
-            Server(boost::asio::io_service& io_service, const std::string& secret, uint16_t port, const std::string& filePath, std::stop_token token, PluginLogger& logger);
+            Server(boost::asio::io_service& io_service, const std::string& secret, uint16_t port, const std::string& filePath, std::stop_token token, PluginLogger& logger, STG::Users* users);
+            int findUser(const RadProto::Packet& packet);
```
madf: I don't think `findUser` should be public.

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -48,6 +51,13 @@ RadProto::Packet Server::makeResponse(const RadProto::Packet& request)
     std::vector<uint8_t> vendorValue {0, 0, 0, 3};
     vendorSpecific.push_back(RadProto::VendorSpecific(m_dictionaries.vendorCode("Dlink"), m_dictionaries.vendorAttributeCode("Dlink", "Dlink-User-Level"), vendorValue));
 
+    if (findUser(request))
```
madf: `findUser` returns `int`, but you use it like `bool`. Consider returning `bool` from `findUser`.

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -48,6 +51,13 @@ RadProto::Packet Server::makeResponse(const RadProto::Packet& request)
...
+    {
+        m_logger("Error findUser\n");
+        printfd(__FILE__, "Error findUser\n");
+        return RadProto::Packet(RadProto::ACCESS_REJECT, request.id(), request.auth(), attributes, vendorSpecific);
```
madf: Reject packet should not have all those attributes.

madf: Same below.

madf (changes requested):

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -1,15 +1,20 @@
...
 #include "stg/common.h"
 #include <cstring>
 #include <functional>
+#include <cstdint> //uint8_t, uint32_t
 
 using STG::Server;
 using boost::system::error_code;
 
-Server::Server(boost::asio::io_service& io_service, const std::string& secret, uint16_t port, const std::string& filePath, std::stop_token token, PluginLogger& logger)
+Server::Server(boost::asio::io_service& io_service, const std::string& secret, uint16_t port, const std::string& filePath, std::stop_token token, PluginLogger& logger, STG::Users* users)
```
madf: `Server` is in the `STG` namespace, so just `Users`, not `STG::Users`.

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -48,10 +53,21 @@ RadProto::Packet Server::makeResponse(const RadProto::Packet& request)
     std::vector<uint8_t> vendorValue {0, 0, 0, 3};
     vendorSpecific.push_back(RadProto::VendorSpecific(m_dictionaries.vendorCode("Dlink"), m_dictionaries.vendorAttributeCode("Dlink", "Dlink-User-Level"), vendorValue));
 
-    if (request.type() == RadProto::ACCESS_REQUEST)
-        return RadProto::Packet(RadProto::ACCESS_ACCEPT, request.id(), request.auth(), attributes, vendorSpecific);
+    std::vector<RadProto::Attribute*> attributesEmpty;
+    std::vector<RadProto::VendorSpecific> vendorSpecificEmpty;
 
-    return RadProto::Packet(RadProto::ACCESS_REJECT, request.id(), request.auth(), attributes, vendorSpecific);
+    if (request.type() == RadProto::ACCESS_REQUEST)
```
madf: In cases like this:
```c++
if (condition)
{
    // A lot of code
    return something;
}
return someOtherThing;
```
consider turning inside out, like this:
```c++
if (!condition)
    return someOtherThing;
    
// A lot of code
return something;
```

It looks cleaner, it is easier to understand and it can reduce cyclomatic complexity.

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -48,10 +53,21 @@ RadProto::Packet Server::makeResponse(const RadProto::Packet& request)
...
+    {
+        if (findUser(request))
+            return RadProto::Packet(RadProto::ACCESS_ACCEPT, request.id(), request.auth(), attributes, vendorSpecific);
+        else
```
madf: You do not need `else` after `return`.

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -48,10 +53,21 @@ RadProto::Packet Server::makeResponse(const RadProto::Packet& request)
...
+        {
+            m_logger("Error findUser\n");
+            printfd(__FILE__, "Error findUser\n");
+            return RadProto::Packet(RadProto::ACCESS_REJECT, request.id(), request.auth(), attributesEmpty, vendorSpecificEmpty);
```
madf: If you need an empty vector, you don't have to create a dedicated variable. You can construct it in-place:
```c++
return RadProto::Packet(RadProto::ACCESS_REJECT, request.id(), request.auth(), {}, {});
```

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -48,10 +53,21 @@ RadProto::Packet Server::makeResponse(const RadProto::Packet& request)
...
+        }
+    }
+    return RadProto::Packet(RadProto::ACCESS_REJECT, request.id(), request.auth(), attributesEmpty, vendorSpecificEmpty);
```
madf: Same as above.

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -84,5 +100,41 @@ void Server::handleReceive(const error_code& error, const std::optional<RadProto
         printfd(__FILE__, "Error asyncReceive: the request packet is missing\n");
         return;
     }
+
     m_radius.asyncSend(makeResponse(*packet), source, [this](const auto& ec){ handleSend(ec); });
 }
+
+bool Server::findUser(const RadProto::Packet& packet)
+{
+    STG::User* user;
```
madf: `Server` is in the `STG` namespace, you don't need to prefix `User` with `STG::`.

madf: Also, please move it close to the point of use (`FindByName` call, I believe) and initialize with `nullptr`.

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -84,5 +100,41 @@ void Server::handleReceive(const error_code& error, const std::optional<RadProto
...
+    printfd(__FILE__, "User '%s' FOUND!\n", user->GetLogin().c_str());
+
+    if (password != user->GetProperties().password.Get())
+    {
+        m_logger("Password user is incorrect. %s", password.c_str());
+        printfd(__FILE__, "Password user is incorrect.\n", password.c_str());
+        return false;
+    }
+
+    printfd(__FILE__, "User FOUND!\n");
```
madf: You've already said that the user is found, no need to repeat it twice.

madf (changes requested):

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -118,6 +111,7 @@ bool Server::findUser(const RadProto::Packet& packet)
             password = attribute->toString();
     }
 
+    User* user(nullptr);
```
madf: `User* user = nullptr;`

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -48,10 +53,15 @@ RadProto::Packet Server::makeResponse(const RadProto::Packet& request)
...
 
-    if (request.type() == RadProto::ACCESS_REQUEST)
+    if (request.type() != RadProto::ACCESS_REQUEST)
+        return RadProto::Packet(RadProto::ACCESS_REJECT, request.id(), request.auth(), {}, {});
+
+    if (findUser(request))
         return RadProto::Packet(RadProto::ACCESS_ACCEPT, request.id(), request.auth(), attributes, vendorSpecific);
 
-    return RadProto::Packet(RadProto::ACCESS_REJECT, request.id(), request.auth(), attributes, vendorSpecific);
+    m_logger("Error findUser\n");
```
madf: We do not need to log it here, because we do it in the `findUser`.

madf (changes requested):

projects/stargazer/plugins/other/radius/radius.h
```diff
@@ -74,7 +77,7 @@ namespace STG
             bool m_running;
 
             std::jthread m_thread;
-
+            STG::Users* m_users;
```
madf: You don't need `STG::`.

projects/stargazer/plugins/other/radius/server.h
```diff
@@ -11,20 +11,24 @@
 
 namespace STG
 {
+    class Users;
+
     class Server
     {
         public:
-            Server(boost::asio::io_service& io_service, const std::string& secret, uint16_t port, const std::string& filePath, std::stop_token token, PluginLogger& logger);
+            Server(boost::asio::io_service& io_service, const std::string& secret, uint16_t port, const std::string& filePath, std::stop_token token, PluginLogger& logger, STG::Users* users);
```
madf: You do not need `STG::`, just `Users`.

projects/stargazer/plugins/other/radius/server.h
```diff
@@ -11,20 +11,24 @@
...
             RadProto::Packet makeResponse(const RadProto::Packet& request);
+            bool findUser(const RadProto::Packet& packet);
             void handleReceive(const boost::system::error_code& error, const std::optional<RadProto::Packet>& packet, const boost::asio::ip::udp::endpoint& source);
             void handleSend(const boost::system::error_code& ec);
             void start();
             void startReceive();
 
             RadProto::Socket m_radius;
             RadProto::Dictionaries m_dictionaries;
+            STG::Users* m_users;
```
madf: You do not need `STG::`.

madf (changes requested):

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -84,5 +93,38 @@ void Server::handleReceive(const error_code& error, const std::optional<RadProto
...
+            login = attribute->toString();
+
+        if (attribute->type() == RadProto::USER_PASSWORD)
+            password = attribute->toString();
+    }
+
+    User* user = nullptr;
+    if (m_users->FindByName(login, &user))
+    {
+        m_logger("User's connect failed: user '%s' not found.", login.c_str());
```
madf: 'User's connect failed' is not what happens here. Just "User '%s' not found" would be ok, I think.

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -84,5 +93,38 @@ void Server::handleReceive(const error_code& error, const std::optional<RadProto
...
+        printfd(__FILE__, "User '%s' NOT found!\n", login.c_str());
+        return false;
+    }
+
+    printfd(__FILE__, "User '%s' FOUND!\n", user->GetLogin().c_str());
+
+    if (password != user->GetProperties().password.Get())
+    {
+        m_logger("Password user is incorrect. %s", password.c_str());
```
madf: Not 'Password user', but 'User's password'.

madf: Same below.

madf (changes requested):

PR #12, opened 2025-05-02
Radiusю

HelenMamontova:
Search for parameters auth and send, get value of the parameter send added.


projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -46,14 +46,27 @@ int RAD_SETTINGS::ParseSettings(const ModuleSettings & s)
...
-    }
 
     pv.param = "Dictionaries";
     pvi = std::find(s.moduleParams.begin(), s.moduleParams.end(), pv);
     if (pvi != s.moduleParams.end() && !pvi->value.empty())
         m_dictionaries = pvi->value[0];
+
+    pv.param = "auth";
+    pvi = std::find(s.moduleParams.begin(), s.moduleParams.end(), pv);
+    if (pvi != s.moduleParams.end() )
```
madf: Extra whitespace at the end.

projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -46,14 +46,27 @@ int RAD_SETTINGS::ParseSettings(const ModuleSettings & s)
...
+    {
+        pv.param = "send";
+        std::string sendValue;
```
madf: You do not need this variable, see below.

projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -46,14 +46,27 @@ int RAD_SETTINGS::ParseSettings(const ModuleSettings & s)
...
+        std::vector<ParamValue> sectionsParams = pvi->sections;
```
madf: You do not need to store sections in the additional variable, see below.

projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -46,14 +46,27 @@ int RAD_SETTINGS::ParseSettings(const ModuleSettings & s)
...
+        auto pva = std::find(sectionsParams.begin(), sectionsParams.end(), pv);
```
madf: Use `sections` directly:
```c++
auto pva = std::find(pvi->sections.begin(), pvi->sections.end(), pv);
```

projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -46,14 +46,27 @@ int RAD_SETTINGS::ParseSettings(const ModuleSettings & s)
...
+        if (pva != sectionsParams.end() && !pva->value.empty())
+        {
+            sendValue = pva->value[0];
+            printfd(__FILE__, "ParseSettings Value of send: '%s'\n", sendValue.c_str());
```
madf: Consider `pva->value[0].c_str()` instead of `sendValue`.

madf (changes requested):

projects/stargazer/plugins/other/radius/radius.h
```diff
@@ -20,11 +20,23 @@ namespace STG
 
     class Users;
 
+    struct AttrValue
+    {
+        enum class Sign
```
madf: Let's call it `Type`.

projects/stargazer/plugins/other/radius/radius.h
```diff
@@ -20,11 +20,23 @@ namespace STG
...
+        {
+            NOT_VALUE,
```
madf: Here would be `PARAM_NAME`

projects/stargazer/plugins/other/radius/radius.h
```diff
@@ -20,11 +20,23 @@ namespace STG
...
+            IS_VALUE
```
madf: And here would be `VALUE`.

projects/stargazer/plugins/other/radius/radius.h
```diff
@@ -20,11 +20,23 @@ namespace STG
 
     class Users;
 
+    struct AttrValue
```
madf: Let's put it into the `RAD_SETTINGS` class.

projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -1,19 +1,64 @@
...
+    boost::char_separator<char> sep(",");
+
+    tokenizer tokens(fieldSendAttr, sep);
+
+    AttrValue attrValue;
+    std::vector<std::pair<std::string, AttrValue>> keyValuePairs;
+
+    for (const auto& token : tokens)
+    {
+        boost::char_separator<char> sp(" =");
```
madf: What is there there is no a whitespace before the '='?

projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -1,19 +1,64 @@
...
+        tokenizer tok(token, sp);
+
+        std::vector<std::string> parsedSendAttr;
+        for (const auto& t : tok)
+            parsedSendAttr.push_back(t);
+
+        std::string key = parsedSendAttr[0];
```
madf: What if `parsedSendAttr` is empty?

projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -1,19 +1,64 @@
...
+        std::string valueName = parsedSendAttr[1];
```
madf: What if `parsedSendAttr` has only one element?

projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -1,19 +1,64 @@
...
+
+        if (valueName[0] == '\'')
+        {
+            valueName.erase(0, 1);
+            valueName.erase(valueName.length() - 1, 1);
```
madf, line 72: What if the last symbol of `valueName` is not a quotation mark?

projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -1,19 +1,64 @@
...
+
+            attrValue.value = valueName;
+            attrValue.sign = AttrValue::Sign::IS_VALUE;
+            keyValuePairs.emplace_back(key, attrValue);
```
madf: Consider `keyValuePairs.emplace_back(key, AttrValue{valueName, AttrValue::Sign::IS_VALUE})`. Same below.

madf (changes requested):

projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -1,22 +1,82 @@
...
 }
 
+std::vector<std::pair<std::string, AttrValue>> RAD_SETTINGS::ParseSendAttr(std::string fieldSendAttr)
+{
+    using tokenizer =  boost::tokenizer<boost::char_separator<char>>;
+    boost::char_separator<char> sep(",");
+
+    tokenizer tokens(fieldSendAttr, sep);
+
+    AttrValue attrValue;
```
madf: Unused variable.

projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -1,22 +1,82 @@
...
+    for (const auto& token : tokens)
+    {
+        boost::char_separator<char> sp(" =");
+        tokenizer tok(token, sp);
+
+        std::vector<std::string> parsedSendAttr;
+        for (const auto& t : tok)
+            parsedSendAttr.push_back(t);
+
+        if (parsedSendAttr.empty())
```
madf: Consider the following:
```c++
if (parsedSendAttr.size() != 2)
{
    m_logger("Send attribute specification has an incorrect format: '%s'.", token.c_str());
    ...
    return {};
}

auto type = AttrValue::Type::PARAM_NAME;
if (/*check if parsedSendAttr[1] has quotes*/)
{
    type = AttrValue::Type::Value;
    /*strip quotes from the parsedSendAttr[1]*/
}
else if (/*check if the quotes are unpaired*/)
{
    /*log error*/
    return {};
}
keyValuePairs.emplace_back(parsedSendAttr[0], AttrValue{parsedSendAttr[1], type});
```

madf (changes requested):

projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -1,22 +1,71 @@
...
 using STG::RAD_SETTINGS;
+using AttrValue = RAD_SETTINGS::AttrValue;
 
 extern "C" STG::Plugin* GetPlugin()
 {
     static RADIUS plugin;
     return &plugin;
 }
 
+std::vector<std::pair<std::string, AttrValue>> RAD_SETTINGS::ParseSendAttr(std::string fieldSendAttr)
```
madf: First, `const std::string&`

Second, let's call the argument `val` or `value`.

projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -1,22 +1,71 @@
...
+{
+    using tokenizer =  boost::tokenizer<boost::char_separator<char>>;
+    boost::char_separator<char> sep(",");
+
+    tokenizer tokens(fieldSendAttr, sep);
+
+    std::vector<std::pair<std::string, AttrValue>> keyValuePairs;
```
madf: Let's call it `res`. It's the result of the function.

projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -1,22 +1,71 @@
...
+
+    for (const auto& token : tokens)
+    {
+        boost::char_separator<char> sp(" =");
+        tokenizer tok(token, sp);
+
+        std::vector<std::string> parsedSendAttr;
```
madf: Let's call it `kv` or `keyValue`.

projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -1,22 +1,71 @@
...
+        if (parsedSendAttr.size() != 2)
+        {
+            m_logger("Send attribute specification has an incorrect format: '%s'.", token.c_str());
+            printfd(__FILE__, "Send attribute specification has an incorrect format: '%s'.", token.c_str());
+            return {};
+        }
+
+        auto type = AttrValue::Type::PARAM_NAME;
+        std::string valueName = parsedSendAttr[1];
+        if (valueName[0] == '\'' && valueName[valueName.length() - 1] == '\'')
```
madf: Consider using `valueName.back()` instead of `valueName[valueName.length() - 1]`. You can also use `front()` for accessing the first char, for symmetry. But it's up to you.

projects/stargazer/plugins/other/radius/radius.h
```diff
@@ -25,8 +25,21 @@ namespace STG
...
+                    PARAM_NAME,
+                    VALUE
+                };
+                std::string value;
+                Type type;
+            };
+
             const std::string & GetStrError() const { return m_errorStr; }
             int ParseSettings(const ModuleSettings & s);
+            std::vector<std::pair<std::string, AttrValue>> ParseSendAttr(std::string fieldSendAttr);
```
madf: Let's make it private.

projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -46,14 +95,29 @@ int RAD_SETTINGS::ParseSettings(const ModuleSettings & s)
...
+    pvi = std::find(s.moduleParams.begin(), s.moduleParams.end(), pv);
+    if (pvi != s.moduleParams.end())
+    {
+        pv.param = "send";
+        auto pva = std::find(pvi->sections.begin(), pvi->sections.end(), pv);
+        if (pva != pvi->sections.end() && !pva->value.empty())
+        {
+            printfd(__FILE__, "ParseSettings Value of send: '%s'\n", pva->value[0].c_str());
+
+            std::vector<std::pair<std::string, AttrValue>> keyValuePairs = ParseSendAttr(pva->value[0]);
```
madf: This variable must be a `RAD_SETTINGS` private member, with a public accessor function.

madf (changes requested):

projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -18,48 +18,46 @@ extern "C" STG::Plugin* GetPlugin()
...
         }
-        else if ((valueName[0] == '\'' && valueName[valueName.length() - 1] != '\'') || (valueName[0] != '\'' && valueName[valueName.length() - 1] == '\''))
+        else if ((valueName.front() == '\'' && valueName.back() != '\'') || (valueName.front() != '\'' && valueName.back() == '\''))
         {
             m_logger("Error ParseSendAttr: send attribute parameter value is invalid.\n");
             printfd(__FILE__, "Error ParseSendAttr: send attribute parameter value is invalid.\n");
             return {};
         }
-        keyValuePairs.emplace_back(parsedSendAttr[0], AttrValue{valueName, type});
+        m_res.emplace_back(keyValue[0], AttrValue{valueName, type});
```
madf: You use a private member `m_res` as a container. And later you return it by value. And the returned value you assign back to the `m_res`.

Here is what happens. First, you fill the `m_res` with parsed data. Now `m_res` contains all necessary information and no further actions are needed. When you return the `m_res` by value, you make a copy of it. And now you have two identical copies of the information. And then you assign the copy back to the original. This assignment does nothing. It's like the following:

```c++
int a = 5;
int b = a;
a = b;
```

One solution would be to return `void`. You've done all the necessary work inside the function, nothing else is needed. However, there are flaws doing thing this way. When the function works, it changes the value of the `m_res`. For now it's nothing, because it is initially empty. However, if we add a reload capability later, it may contain correct data. And you will gradually change it. It will introduce a race condition and you would have to guard each `emplace_back` with a mutex. What's worse, if there is an issue with the input data, the parsing may fail. Buth the original value of the `m_res` would be damaged.

The better way is to use a local container `res`, not a class member. And fill it with values. And return it by value. And in the end, assign the `m_res` to the returned value.

projects/stargazer/plugins/other/radius/radius.h
```diff
@@ -37,19 +37,22 @@ namespace STG
...
-            std::vector<std::pair<std::string, AttrValue>> ParseSendAttr(std::string fieldSendAttr);
+            const std::string& GetStrError() const { return m_errorStr; }
+            int ParseSettings(const ModuleSettings& s);
 
             uint16_t GetPort() const { return m_port; }
-            const std::string & GetDictionaries() const { return m_dictionaries; }
-            const std::string & GetSecret() const { return m_secret; }
+            const std::string& GetDictionaries() const { return m_dictionaries; }
+            const std::string& GetSecret() const { return m_secret; }
+            const std::vector<std::pair<std::string, AttrValue>>& GetRes() const { return m_res; }
```
madf: `GetRes` and `m_res` are bad names. The mean nothing. This field represents the contents of the `send` field in the config, so let's call it `GetSend` or `GetSendPairs`, and `m_send` or `m_sendPairs` respectively.

madf (changes requested):

projects/stargazer/inst/linux/etc/stargazer/conf-available.d/mod_radius.conf
```diff
@@ -17,4 +17,8 @@
     # Default: 1812
     # Port = 1812
 
+    <auth>
+        send = "Framed-IP-Address = currip, Framed-Protocol = \'PPP\'"
+        match = "Framed-IP-Address = currip, Framed-Protocol = \'PPP\'"
```
madf: This 'match' example does not make much sense. Let's use something like this:
```
match="Calling-Station-Id=userdata0"
```

madf (changes requested):

projects/stargazer/inst/linux/etc/stargazer/conf-available.d/mod_radius.conf
```diff
@@ -19,6 +19,11 @@
 
     <auth>
         send = "Framed-IP-Address = currip, Framed-Protocol = \'PPP\'"
```
madf: Let's use "Service-Type = 'Framed-User'" instead of "Framed-Protocol = 'PPP'" here.

projects/stargazer/inst/linux/etc/stargazer/conf-available.d/mod_radius.conf
```diff
@@ -19,6 +19,11 @@
...
-        match = "Framed-IP-Address = currip, Framed-Protocol = \'PPP\'"
+        match = "Calling-Station-Id = userdata0"
     </auth>
+
+    <autz>
+        send = "Framed-IP-Address = currip, Framed-Protocol = \'PPP\'"
```
madf: And use "Service-Type = 'Login-User'" instead of "Framed-Protocol = 'PPP'" here.

projects/stargazer/inst/linux/etc/stargazer/conf-available.d/mod_radius.conf
```diff
@@ -19,6 +19,11 @@
...
+        match = "Calling-Station-Id = userdata0"
```
madf: And use `match  = "User-Name = login, User-Password = password" here.

projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -18,7 +18,7 @@ extern "C" STG::Plugin* GetPlugin()
     return &plugin;
 }
 
-std::vector<std::pair<std::string, AttrValue>> RAD_SETTINGS::ParseAuthAttr(const std::string& value, const std::string& paramName)
+std::vector<std::pair<std::string, AttrValue>> RAD_SETTINGS::ParseAuthAutzAttr(const std::string& value, const std::string& paramName)
```
madf: Let's call it `ParseRules`.

projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -109,9 +109,9 @@ int RAD_SETTINGS::ParseSettings(const ModuleSettings & s)
         auto pva = std::find(pvi->sections.begin(), pvi->sections.end(), pv);
         if (pva != pvi->sections.end() && !pva->value.empty())
         {
-            printfd(__FILE__, "ParseSettings Value of send: '%s'\n", pva->value[0].c_str());
+            printfd(__FILE__, "ParseSettings Value of auth/send: '%s'\n", pva->value[0].c_str());
```
madf: I think, we no longer need this print.

projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -109,9 +109,9 @@ int RAD_SETTINGS::ParseSettings(const ModuleSettings & s)
...
 
-            m_sendPairs = ParseAuthAttr(pva->value[0], pv.param);
+            m_sendPairs = ParseAuthAutzAttr(pva->value[0], pv.param);
 
             for (const auto& at : m_sendPairs)
                 printfd(__FILE__, "Key: '%s', Value: '%s', Type: %d\n", at.first.c_str(), at.second.value.c_str(), at.second.type);
```
madf: Let's make a function, something like this:
```c++
std::string showRules(const std::vector<std::pair<std::string, AttrValue>>)
{
    // ...
}
```
Which will combine the pairs back to the original form, and use it like this:
```c++
printfd(__FILE__, "auth.match = \"%s\"\n", showRules(m_matchPairs).c_str());
...
```

projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -121,9 +121,38 @@ int RAD_SETTINGS::ParseSettings(const ModuleSettings & s)
...
+    pvi = std::find(s.moduleParams.begin(), s.moduleParams.end(), pv);
+    if (pvi != s.moduleParams.end())
+    {
+        pv.param = "send";
+        auto pva = std::find(pvi->sections.begin(), pvi->sections.end(), pv);
+        if (pva != pvi->sections.end() && !pva->value.empty())
+        {
+            printfd(__FILE__, "ParseSettings Value of autz/send: '%s'\n", pva->value[0].c_str());
+
+            m_sendPairs = ParseAuthAutzAttr(pva->value[0], pv.param);
```
madf: Here you overwrite the `m_sendPairs`, which contains the value from the 'auth' section. You need additional pair of `m_sendPairs` and `m_matchPairs` for the 'autz' section.

madf (changes requested):

projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -1,22 +1,127 @@
...
+std::string RAD_SETTINGS::ShowRules(const std::vector<std::pair<std::string, AttrValue>> mvector)
+{
+    std::string strRules;
+    size_t i = 0;
+    for (const auto& at : mvector)
+    {
+        if (at.second.type == AttrValue::Type::PARAM_NAME)
+            strRules.append(at.first + " = " + at.second.value);
+        else
+            strRules.append(at.first + " = " + "'" + at.second.value + "'");
```
madf: You can combine `" = " + "'"` to `" = '"`

projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -1,22 +1,127 @@
...
+            return {};
+        }
+        res.emplace_back(keyValue[0], AttrValue{valueName, type});
+    }
+    return res;
+}
+
+std::string RAD_SETTINGS::ShowRules(const std::vector<std::pair<std::string, AttrValue>> mvector)
+{
+    std::string strRules;
```
madf: Let's just call it `res`, as "result".

projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -1,22 +1,127 @@
...
+    size_t i = 0;
+    for (const auto& at : mvector)
+    {
+        if (at.second.type == AttrValue::Type::PARAM_NAME)
+            strRules.append(at.first + " = " + at.second.value);
+        else
+            strRules.append(at.first + " = " + "'" + at.second.value + "'");
+
+        if (i != mvector.size() - 1)
```
madf: This `i` counter is really needed once here - at the last element. I propose a slightly modified scheme:
```c++
bool first = true;
for (const auto& at: mvector)
{
    if (first)
        first = false;
    else
        res += ", ";
        
    // The rest of the code...
}
```

projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -1,22 +1,127 @@
...
+            m_logger("Error ParseRules: '%s' attribute parameter value is invalid.\n", paramName.c_str());
+            printfd(__FILE__, "Error ParseRules: '%s' attribute parameter value is invalid.\n", paramName.c_str());
+            return {};
+        }
+        res.emplace_back(keyValue[0], AttrValue{valueName, type});
+    }
+    return res;
+}
+
+std::string RAD_SETTINGS::ShowRules(const std::vector<std::pair<std::string, AttrValue>> mvector)
```
madf: Missed `&`.

madf: Also let's call it `attributes` or `attrs`. "mvector" says nothing about its nature.

projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -1,22 +1,127 @@
...
+            strRules.append(at.first + " = " + "'" + at.second.value + "'");
+
+        if (i != mvector.size() - 1)
+            strRules.append(", ");
+        ++i;
+    }
+    return strRules;
+}
+
+void RAD_SETTINGS::MakeKeyValuePairs(const ModuleSettings & s, ParamValue pv, const std::string& paramName)
```
madf: I propose to make it more generic. First, let's make a type:
```c++
struct ASection
{
    using Pairs = std::vector<std::pair<std::string, AttrValue>>;
    Pairs match;
    Pairs send;
};
```

In the `RAD_SETTINGS` use it like this:
```c++
class RAD_SETTINGS
{
    public:
        // ...
        const ASection& getAuth() const { return m_auth; }
        const ASection& getAutz() const { return m_autz; }
        // ...
    private:
        // ...
        ASection m_auth;
        ASection m_autz;
        // ...
};
```

And this function will be:
```c++
ASection parseASection(const std::vector<ParamValue>& conf)
{
    ASection res;
    const auto mit = std::find(conf.begin(), conf.end(), ParamValue("match", {}));
    if (mit != conf.end())
        res.match = ParseRules(mit->value[0], mit->param);
    const auto sit = std::find(conf.begin(), conf.end(), ParamValue("send", {}));
    if (sit != conf.end())
        res.send = ParseRules(sit->value[0], sit->param);
}
```

And then in the `ParseSettings`:
```c++
const auto authIt = std::find(s.moduleParams.begin(), s.moduleParams.end(), ParamValue("auth", {}));
if (authIt != s.moduleParams.end())
    m_auth = parseASection(authIt->sections);
const auto autzIt = std::find(s.moduleParams.begin(), s.moduleParams.end(), ParamValue("autz", {}));
if (autzIt != s.moduleParams.end())
    m_autz = parseASection(autzIt->sections);
printfd(__FILE__, " auth.match = \"%s\"\n", ShowRules(m_auth.match).c_str());
printfd(__FILE__, " auth.send = \"%s\"\n", ShowRules(m_auth.send).c_str());
printfd(__FILE__, " autz.match = \"%s\"\n", ShowRules(m_autz.match).c_str());
printfd(__FILE__, " autz.send = \"%s\"\n", ShowRules(m_autz.send).c_str());
```

madf (changes requested):

projects/stargazer/plugins/other/radius/radius.h
```diff
@@ -47,27 +77,27 @@ namespace STG
...
 
-            const std::string & GetStrError() const override { return m_errorStr; }
+            const std::string& GetStrError() const override { return m_errorStr; }
             std::string GetVersion() const override;
 
             uint16_t GetStartPosition() const override { return 0; }
             uint16_t GetStopPosition() const override { return 0; }
 
-            int SendMessage(const Message & msg, uint32_t ip) const override { return 0; }
+            int SendMessage(const Message& msg, uint32_t ip) const override { return 0; }
```
madf:
```
/tmp/stg/projects/stargazer/plugins/other/radius/radius.h: In member function ‘virtual int STG::RADIUS::SendMessage(const STG::Message&, uint32_t) const’:
/tmp/stg/projects/stargazer/plugins/other/radius/radius.h:95:44: warning: unused parameter ‘msg’ [-Wunused-parameter]
   95 |             int SendMessage(const Message& msg, uint32_t ip) const override { return 0; }
      |                             ~~~~~~~~~~~~~~~^~~
/tmp/stg/projects/stargazer/plugins/other/radius/radius.h:95:58: warning: unused parameter ‘ip’ [-Wunused-parameter]
   95 |             int SendMessage(const Message& msg, uint32_t ip) const override { return 0; }

```

projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -1,22 +1,104 @@
...
 extern "C" STG::Plugin* GetPlugin()
 {
     static RADIUS plugin;
     return &plugin;
 }
 
+std::vector<std::pair<std::string, AttrValue>> RAD_SETTINGS::ParseRules(const std::string& value, const std::string& paramName)
+{
+    using tokenizer =  boost::tokenizer<boost::char_separator<char>>;
+    boost::char_separator<char> sep(",");
```
madf:
```
/tmp/stg/projects/stargazer/plugins/other/radius/radius.cpp:25:5: warning: variable 'sep' of type 'boost::char_separator<char>' can be declared 'const' [misc-const-correctness]
   25 |     boost::char_separator<char> sep(",");
      |     ^
      |                                 const 
```

projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -1,22 +1,104 @@
...
+
+    tokenizer tokens(value, sep);
```
madf:
```
/tmp/stg/projects/stargazer/plugins/other/radius/radius.cpp:27:5: warning: variable 'tokens' of type 'tokenizer' (aka 'tokenizer<boost::char_separator<char>>') can be declared 'const' [misc-const-correctness]
   27 |     tokenizer tokens(value, sep);
      |     ^
      |               const
```

projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -1,22 +1,104 @@
...
+
+    std::vector<std::pair<std::string, AttrValue>> res;
+    for (const auto& token : tokens)
+    {
+        boost::char_separator<char> sp(" =");
```
madf:
```
/tmp/stg/projects/stargazer/plugins/other/radius/radius.cpp:32:9: warning: variable 'sp' of type 'boost::char_separator<char>' can be declared 'const' [misc-const-correctness]
   32 |         boost::char_separator<char> sp(" =");
      |         ^
      |                                     const
```

projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -1,22 +1,104 @@
...
+        tokenizer tok(token, sp);
```
madf:
```
/tmp/stg/projects/stargazer/plugins/other/radius/radius.cpp:33:9: warning: variable 'tok' of type 'tokenizer' (aka 'tokenizer<boost::char_separator<char>>') can be declared 'const' [misc-const-correctness]
   33 |         tokenizer tok(token, sp);
      |         ^
      |                   const
```

projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -1,22 +1,104 @@
...
+            m_logger("Error ParseRules: '%s' attribute parameter value is invalid.\n", paramName.c_str());
+            printfd(__FILE__, "Error ParseRules: '%s' attribute parameter value is invalid.\n", paramName.c_str());
+            return {};
+        }
+        res.emplace_back(keyValue[0], AttrValue{valueName, type});
+    }
+    return res;
+}
+
+std::string RAD_SETTINGS::ShowRules(const std::vector<std::pair<std::string, AttrValue>>& attributes)
```
madf: This function does not use anything from `RAD_SETTINGS`. Consider making it a free function.

madf (changes requested):

projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -19,18 +19,37 @@ extern "C" STG::Plugin* GetPlugin()
     return &plugin;
 }
 
+std::string ShowRules(const std::vector<std::pair<std::string, AttrValue>>& attributes)
```
madf: Please put it into the anonymous namespace.

projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -19,18 +19,37 @@ extern "C" STG::Plugin* GetPlugin()
...
+{
+    std::string result;
+    bool first = true;
+    for (const auto& at : attributes)
+    {
+        if (first)
```
madf: I think we can re-work this coma placement code. What do you think about this:
```c++
if (!result.empty())
    result += ", ";
```

projects/stargazer/plugins/other/radius/radius.h
```diff
@@ -47,27 +76,27 @@ namespace STG
             RADIUS& operator=(const RADIUS&) = delete;
 
             void SetUsers(Users* u) { m_users = u; }
```
madf: This function must be marked `override`.

projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -1,22 +1,104 @@
 #include "radius.h"
 #include "radproto/error.h"
 #include "stg/common.h"
+#include <boost/tokenizer.hpp>
 
+#include <vector>
+#include <utility>
 #include <iterator>
```
madf, line 8: Add `#include <algorithm>` - you use `std::find` from there.

madf (changes requested):

PR #13, opened 2025-08-28
Sending attributes from <auth>/send section.

HelenMamontova:
Class RAD_SETTINGS is transferred to separate files .h / .cpp. The makeResponse function changed to send attributes configured in the <auth>/send section. Function makeAttributes added.


projects/stargazer/inst/linux/etc/stargazer/conf-available.d/mod_radius.conf
```diff
@@ -18,7 +18,7 @@
     # Port = 1812
 
     <auth>
-        send = "Framed-IP-Address = currip, Service-Type = \'Framed-User\'"
+        send = "User-Name = login, Framed-IP-Address = currip, Service-Type = \'Framed-User\', Callback-Number = \'987654abc\'"
```
madf: Let's keep the original value. It makes more sense.

projects/stargazer/plugins/CMakeLists.txt
```diff
@@ -15,7 +15,8 @@ if ( BUILD_MOD_RADIUS )
     find_package ( Boost REQUIRED )
 
     add_library ( mod_radius MODULE other/radius/radius.cpp
-                                    other/radius/server.cpp)
+                                    other/radius/server.cpp
+                                    other/radius/rad_settings.cpp)
```
madf: How about just `settings.cpp`?

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -4,17 +4,23 @@
...
 using STG::Server;
+using STG::User;
 using boost::system::error_code;
 
-Server::Server(boost::asio::io_context& io_context, const std::string& secret, uint16_t port, const std::string& filePath, std::stop_token token, PluginLogger& logger, Users* users)
+Server::Server(boost::asio::io_context& io_context, const std::string& secret, uint16_t port, const std::string& filePath, std::stop_token token, PluginLogger& logger, Users* users, const RAD_SETTINGS& radSettings)
     : m_radius(io_context, secret, port),
       m_dictionaries(filePath),
       m_users(users),
+      m_radSettings(radSettings),
```
madf: Let's call them `settings` and `m_settings`.

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -37,27 +43,78 @@ void Server::startReceive()
     m_radius.asyncReceive([this](const auto& error, const auto& packet, const boost::asio::ip::udp::endpoint& source){ handleReceive(error, packet, source); });
 }
 
-RadProto::Packet Server::makeResponse(const RadProto::Packet& request)
+std::vector<RadProto::Attribute*> Server::makeAttributes(const User* user)
```
madf, line 46: I think this code is derived from the `async-radius` library, isn't it? If so, let's just expose it in the lib and use here.

madf (changes requested):

projects/stargazer/plugins/other/radius/config.cpp
```diff
@@ -0,0 +1,147 @@
...
+#include <algorithm>
+#include <iostream>
+
+using STG::Config;
+using AttrValue = Config::AttrValue;
+using ASection = Config::ASection;
+
+namespace
+{
+    std::string ShowRules(const std::vector<std::pair<std::string, AttrValue>>& attributes)
```
madf: `attributes` -> `rules`

Because these are rules, not attributes.

projects/stargazer/plugins/other/radius/config.cpp
```diff
@@ -0,0 +1,147 @@
...
+            else
+                result.append(at.first + " = '" + at.second.value + "'");
+        }
+        return result;
+    }
+}
+
+std::vector<std::pair<std::string, AttrValue>> Config::ParseRules(const std::string& value, const std::string& paramName)
+{
+    using tokenizer =  boost::tokenizer<boost::char_separator<char>>;
```
madf: Double whitespace after `=`.

projects/stargazer/plugins/other/radius/config.cpp
```diff
@@ -0,0 +1,147 @@
...
+    const boost::char_separator<char> sep(",");
+
+    const tokenizer tokens(value, sep);
+
+    std::vector<std::pair<std::string, AttrValue>> res;
+    for (const auto& token : tokens)
+    {
+        const boost::char_separator<char> sp(" =");
+        const tokenizer tok(token, sp);
```
madf: Consider using `boost::split` here. It gives you `std::vector<std::string>` right away, no need to copy it manually. See the example at the bottom of https://www.boost.org/doc/libs/latest/doc/html/string_algo/usage.html

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -37,27 +44,44 @@ void Server::startReceive()
...
-    attributes.push_back(new RadProto::Bytes(m_dictionaries.attributeCode("Callback-Number"), bytes));
-    std::vector<uint8_t> chapPassword {'1', '2', '3', '4', '5', '6', '7', '8', '9', 'a', 'b', 'c', 'd', 'e', 'f', 'g' };
-    attributes.push_back(new RadProto::ChapPassword(m_dictionaries.attributeCode("CHAP-Password"), 1, chapPassword));
-
-    std::vector<RadProto::VendorSpecific> vendorSpecific;
-    std::vector<uint8_t> vendorValue {0, 0, 0, 3};
-    vendorSpecific.push_back(RadProto::VendorSpecific(m_dictionaries.vendorCode("Dlink"), m_dictionaries.vendorAttributeCode("Dlink", "Dlink-User-Level"), vendorValue));
-
-    if (request.type() != RadProto::ACCESS_REQUEST)
+    std::string attrName;
```
madf: `attrName`, `attrValue`, `attrType` and `attrCode` are only used inside the cycle. Please move them there.

madf (changes requested):

projects/stargazer/plugins/other/radius/config.cpp
```diff
@@ -2,6 +2,9 @@
 #include "radproto/error.h"
 #include "stg/common.h"
 #include <boost/tokenizer.hpp>
+#include <boost/algorithm/string.hpp>
+#include <boost/algorithm/string/trim.hpp>
+#include <boost/algorithm/string/predicate.hpp>
```
madf: Do you need all three? Isn't it enough to include just `string.hpp`, or two: `trim.hpp` and `predicate.hpp`?

projects/stargazer/plugins/other/radius/config.cpp
```diff
@@ -34,20 +37,18 @@ namespace
...
     for (const auto& token : tokens)
     {
-        const boost::char_separator<char> sp(" =");
-        const tokenizer tok(token, sp);
-
         std::vector<std::string> keyValue;
-        for (const auto& t : tok)
-            keyValue.push_back(t);
+
+        split(keyValue, boost::algorithm::trim_copy_if(token, boost::is_any_of(" \t\r\n\xA0")), boost::is_any_of(" ="), boost::token_compress_on);
```
madf: What is `0xA0`?

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -47,18 +47,20 @@ void Server::startReceive()
...
     for (const auto& at : m_config.getAuth().send)
     {
+        std::string attrValue;
+
         if (at.second.type == Config::AttrValue::Type::PARAM_NAME)
             attrValue = user->GetParamValue(at.second.value);
         else
             attrValue = at.second.value;
 
+        std::string attrName;
```
madf:
```cpp
const auto attrName = at.first;
```
Same for others.

madf (changes requested):

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -124,7 +146,7 @@ bool Server::findUser(const RadProto::Packet& packet)
     {
         m_logger("User's password is incorrect. %s", password.c_str());
```
madf: Please do not expose user password in the log. It's a serious security issue.

projects/stargazer/plugins/other/radius/config.cpp
```diff
@@ -0,0 +1,146 @@
...
+        m_port = static_cast<uint16_t>(p);
+    }
+
+    pv.param = "Secret";
+    pvi = std::find(s.moduleParams.begin(), s.moduleParams.end(), pv);
+    if (pvi == s.moduleParams.end() || pvi->value.empty())
+    {
+        m_errorStr = "Parameter \'Secret\' not found.";
+        printfd(__FILE__, "Parameter 'Secret' not found\n");
+        m_secret = "";
```
madf: I guess, you need to return -1 here.

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -37,27 +44,42 @@ void Server::startReceive()
...
+            attributes.push_back(RadProto::Attribute::make(attrCode, attrType, attrValue));
+    }
+    return attributes;
+}
+
+RadProto::Packet Server::makeResponse(const RadProto::Packet& request)
+{
+    const User* user;
+
+    user = findUser(request);
```
madf, line 78: Do not look for the user until you really need it. Consider that case when most incoming traffic is garbage, and it won't pass your `request.code() != RadProto::ACCESS_REQUEST` check. You do not need the user at this moment. It would simply burn your CPU for unnecessary lookups. Move it down, to the place where you use it.

projects/stargazer/plugins/other/radius/server.cpp
madf, line 108: Shouldn't you do something here? E.g. ``startReceive` and return?

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -124,7 +146,7 @@ bool Server::findUser(const RadProto::Packet& packet)
     {
         m_logger("User's password is incorrect. %s", password.c_str());
         printfd(__FILE__, "User's password is incorrect.\n", password.c_str());
```
madf: `password.c_str()` is unused. And shouldn't be. Please remove.

projects/stargazer/plugins/other/radius/radius.h
madf: Please make it private. It is not used anywhere outside the `Run` method.

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -1,20 +1,27 @@
...
 #include "radproto/packet_codes.h"
-#include "radproto/attribute_types.h"
+#include "radproto/attribute_codes.h"
 #include "stg/user.h"
 #include "stg/users.h"
 #include "stg/common.h"
+#include <vector>
+#include <string>
+#include <sstream>
+#include <algorithm> // для std::replace
```
madf: First, do not use Cyrillic. Even for comments. Second, you do not use `std::replace`.

projects/stargazer/plugins/other/radius/radius.cpp
```diff
@@ -3,155 +3,16 @@
...
-    printfd(__FILE__, " auth.send = \"%s\"\n", ShowRules(m_auth.send).c_str());
-    printfd(__FILE__, " autz.match = \"%s\"\n", ShowRules(m_autz.match).c_str());
-    printfd(__FILE__, " autz.send = \"%s\"\n", ShowRules(m_autz.send).c_str());
-
-    return 0;
-}
-
 RADIUS::RADIUS()
     : m_running(false),
       m_users(NULL),
```
madf: Please use `nullptr` instead of `NULL`.

madf (changes requested):

projects/stargazer/plugins/other/radius/config.cpp
```diff
@@ -118,7 +118,7 @@ int Config::ParseSettings(const ModuleSettings & s)
     {
         m_errorStr = "Parameter \'Secret\' not found.";
         printfd(__FILE__, "Parameter 'Secret' not found\n");
-        m_secret = "";
+        m_secret = -1;
```
madf: What? Does it even compile?!

madf (changes requested):

PR #14, opened 2026-04-17
Change function findUser.

HelenMamontova:
The function findUser changed to search for users according to request and match.


projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -57,8 +57,8 @@ std::vector<RadProto::Attribute*> Server::makeAttributes(const User* user)
             attrValue = at.second.value;
 
         const auto attrName = at.first;
-        const auto attrCode = m_dictionaries.attributeCode(attrName);
-        const auto attrType = m_dictionaries.attributeType(attrCode);
+        uint32_t attrCode = m_dictionaries.attributeCode(attrName);
```
madf: Why did you change `const auto` to a non-const specific type?

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -121,32 +121,77 @@ void Server::handleReceive(const error_code& error, const std::optional<RadProto
...
 {
-    std::string login;
-    std::string password;
+    std::vector<std::pair<std::string, std::string>> requestAttr;
+
     for (const auto& attribute : packet.attributes())
     {
-        if (attribute->code() == RadProto::USER_NAME)
-            login = attribute->toString();
+        std::string requestAttrName = m_dictionaries.attributeName(attribute->code());
```
madf: Missing `const`?

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -121,32 +121,77 @@ void Server::handleReceive(const error_code& error, const std::optional<RadProto
...
+        const std::string reqAttrValue = attribute->toString();
+        std::string requestAttrValue;
 
-        if (attribute->code() == RadProto::USER_PASSWORD)
-            password = attribute->toString();
-    }
+        if ((requestAttrType == "integer") && (m_dictionaries.attributeValueFindByName(requestAttrName, reqAttrValue)))
+            requestAttrValue =  std::to_string(m_dictionaries.attributeValueCode(requestAttrName, reqAttrValue));
+        else
+            requestAttrValue = reqAttrValue;
```
madf: You do not need additional `reqAttrValue` and `else` branch, here is a shorter version of your code:
```c++
auto requestAttrValue = attribute->toString();
if (requestAttrType == "integer" && m_dictionaries.attributeValueFindByName(requestAttrName, reqAttrValue))
    requestAttrValue =  std::to_string(m_dictionaries.attributeValueCode(requestAttrName, requestAttrValue));
```

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -121,32 +121,77 @@ void Server::handleReceive(const error_code& error, const std::optional<RadProto
...
 
-    User* user = nullptr;
-    if (m_users->FindByName(login, &user))
-    {
-        m_logger("User '%s' not found.", login.c_str());
-        printfd(__FILE__, "User '%s' NOT found!\n", login.c_str());
-        return nullptr;
+        requestAttr.push_back(std::make_pair(requestAttrName, requestAttrValue));
```
madf:
```c++
requestAttr.emplace_back(requestAttrName, requestAttrValue);
```

madf: Besides, you do not need all attributes from the request, you need only those that are specified in the `match`.

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -121,32 +121,77 @@ void Server::handleReceive(const error_code& error, const std::optional<RadProto
...
+
+            for (const auto& p : requestAttr)
+            {
+                if (matchAttrName != p.first)
+                    continue;
+
+                const std::string reqValue = p.second;
+                const std::string usValue = userAttrValue;
+
+                if (reqValue == usValue)
```
madf: `usValue` is a `userAttrValue`. `userAttrValue` is a `matchAttrValue`. Why do you need two additional variables? Why can't you just
```c++
if (reqValue == matchAttrValue)
```

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -121,32 +121,77 @@ void Server::handleReceive(const error_code& error, const std::optional<RadProto
...
+                {
+                    nextMatchAttr = true;
+                    break; //go to next match attr
+                }
+                nextUser = true;
+                break;//go to next user
+
+            }// end cycle for request
+            if (nextMatchAttr)
+                continue;
```
madf: If `nextMatchAttr` is `true` we continue the cycle by `match`. It it is `false`... we also continue the cycle by `match`! So what's the purpose of this flag?

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -121,32 +121,77 @@ void Server::handleReceive(const error_code& error, const std::optional<RadProto
...
-    }
-    return user;
+        bool nextUser = false;
+
+        for (const auto& at : m_config.getAuth().match)
+        {
+            bool nextMatchAttr = false;
+
+            std::string matchAttrValue;
+            const auto secondValue = at.second.value;
```
madf: `secondValues` is no better than `at.second.value`. Why don't you use just `at.second.value`?

projects/stargazer/plugins/other/radius/server.h
```diff
@@ -15,6 +15,8 @@ namespace STG
     class Users;
     class User;
 
+    using UserPtr = User*;
```
madf: Why do you need this?

madf (changes requested):

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -121,32 +121,75 @@ void Server::handleReceive(const error_code& error, const std::optional<RadProto
...
-    std::string password;
+    std::vector<std::pair<std::string, std::string>> requestAttr;
+
     for (const auto& attribute : packet.attributes())
     {
-        if (attribute->code() == RadProto::USER_NAME)
-            login = attribute->toString();
+        const std::string requestAttrName = m_dictionaries.attributeName(attribute->code());
+
+        for (const auto& at : m_config.getAuth().match)
```
madf: Better swap these two cycles, `for` by `match` and `for` by `attributes`. In the outer cycle you go by `match` attributes, and in the inner cycle you try to find it among `attributes`. And if you don't find it - return `nullptr` immediately, because all attributes from `match` are required.

When the attributes is found, and the `at.second.type` is `PARAM_VALUE`, you can check the value from `at.second` with the attribute value from the `attributes` and if they do not match - return `nullptr` immediately.

And in the resulting vector (currently you call it `requestAttr`, but think of a better name) you can put attribute value to the left, and param name to the right. After that you will only need to go through the vector for every user and check that all user params taken from the right match by value with the left.

madf (changes requested):

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -121,75 +121,62 @@ void Server::handleReceive(const error_code& error, const std::optional<RadProto
...
         {
+            const std::string requestAttrName = m_dictionaries.attributeName(attribute->code());
+
             if (requestAttrName != at.first)
                 continue;
 
-            const std::string requestAttrType = m_dictionaries.attributeType(requestAttrName);
             auto requestAttrValue = attribute->toString();
 
+            if (at.second.type == Config::AttrValue::Type::VALUE && at.second.value != requestAttrValue)
```
madf: Before comparing `at.second.value != requestAttrValue` you should first check if the attribute is integer and make a dictionary lookup by `at.second.value`.

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -121,75 +121,62 @@ void Server::handleReceive(const error_code& error, const std::optional<RadProto
...
+                return nullptr;
+
+            const std::string requestAttrType = m_dictionaries.attributeType(requestAttrName);
+
             if ((requestAttrType == "integer") && (m_dictionaries.attributeValueFindByName(requestAttrName, requestAttrValue)))
                 requestAttrValue =  std::to_string(m_dictionaries.attributeValueCode(requestAttrName, requestAttrValue));
```
madf: Instead of `m_dictionaries.attributeValueCode(requestAttrName, requestAttrValue)` there should be `m_dictionaries.attributeValueCode(requestAttrName, at.second.value)`

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -121,75 +121,62 @@ void Server::handleReceive(const error_code& error, const std::optional<RadProto
...
 
-            requestAttr.emplace_back(requestAttrName, requestAttrValue);
+            valuesForCompare.emplace_back(requestAttrValue, at.second);
```
madf: You should only place `requestAttrValue` in the vector if the `at.second.type == Config::AttrValue::Type::PARAM_NAME`. Values should be handled immediately.

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -121,75 +121,62 @@ void Server::handleReceive(const error_code& error, const std::optional<RadProto
 
 const User* Server::findUser(const RadProto::Packet& packet)
 {
-    std::vector<std::pair<std::string, std::string>> requestAttr;
+    std::vector<std::pair<std::string, STG::Config::AttrValue>> valuesForCompare;
```
madf: The second half of the pair should be `std::string`, it should only contain user param names.

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -121,75 +121,62 @@ void Server::handleReceive(const error_code& error, const std::optional<RadProto
...
+
             if ((requestAttrType == "integer") && (m_dictionaries.attributeValueFindByName(requestAttrName, requestAttrValue)))
                 requestAttrValue =  std::to_string(m_dictionaries.attributeValueCode(requestAttrName, requestAttrValue));
 
-            requestAttr.emplace_back(requestAttrName, requestAttrValue);
+            valuesForCompare.emplace_back(requestAttrValue, at.second);
+            break;
         }
+        if (valuesForCompare.empty())
+            return nullptr;
```
madf: It's not correct. The check must be out of the cycle. And returning `nullptr` is not correct either. Empty vector here means that we checked all the `match` clauses and all of them are correct. Technically it means that any user matches the request (which is wrong, but only in a sense that the `match` is wrong). So I would proceed to user lookup cycle and return first user.

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -121,75 +121,62 @@ void Server::handleReceive(const error_code& error, const std::optional<RadProto
...
 
-        for (const auto& at : m_config.getAuth().match)
+        for (const auto& p : valuesForCompare)
         {
-            std::string matchAttrValue;
+            std::string paramValue;
 
-            if (at.second.type == Config::AttrValue::Type::PARAM_NAME)
-                matchAttrValue = u->GetParamValue(at.second.value);
+            if (p.second.type == Config::AttrValue::Type::PARAM_NAME)
```
madf: `p.second.type` should always be `Config::AttrValue::Type::PARAM_VALUE` here.

madf (changes requested):

projects/stargazer/plugins/other/radius/server.cpp
madf: I meant `PARAM_NAME`

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -121,32 +121,58 @@ void Server::handleReceive(const error_code& error, const std::optional<RadProto
...
+        for (const auto& p : valuesForCompare)
+        {
+            if (u->GetParamValue(p.second) != p.first)
+            {
+                nextUser = true;
+                break;
+            }
+        }
+        if (nextUser)
+            continue;
```
madf: Use of `nextUser` feels somewhat dirty. Consider the following:
```c++
bool allParamsMatch = true;
for (const auto kv : valuesForCompare)
{
    allParamsMatch = allParamsMatch && p.first == u->GetParamValue(p.second);
    if (!allParamsMatch)
        break;
}
if (allParamsMatch)
{
    m_users->CloseSearch(h);
    return u;
}
```

madf (changes requested):

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -121,32 +121,58 @@ void Server::handleReceive(const error_code& error, const std::optional<RadProto
...
+                continue;
+
+            auto requestAttrVal = attribute->toString();
+            const std::string requestAttrType = m_dictionaries.attributeType(requestAttrName);
+            std::string requestAttrValue;
+
+            if (requestAttrType == "integer" && m_dictionaries.attributeValueFindByName(requestAttrName, at.second.value))
+                requestAttrValue =  std::to_string(m_dictionaries.attributeValueCode(requestAttrName, at.second.value));
+            else
+                requestAttrValue = requestAttrVal;
```
madf: That's not quite correct. You have two values to compare: one from the request and one from the 'match' (it it is a value and not a param name). The value from the request always comes numeric. The value from the 'match' can be either numeric or a name from the dictionary. It means that `requestAttrValue` does not change. You take it from the packet and that's all. The value from the 'match' may change, if the attribute type is 'integer'. So you need to store it in some variable to compare later against `requestAttrValue`.

madf (changes requested):

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -121,32 +121,56 @@ void Server::handleReceive(const error_code& error, const std::optional<RadProto
...
 
-    User* user = nullptr;
-    if (m_users->FindByName(login, &user))
-    {
-        m_logger("User '%s' not found.", login.c_str());
-        printfd(__FILE__, "User '%s' NOT found!\n", login.c_str());
-        return nullptr;
+            auto requestAttrValue = attribute->toString();
+
+            std::string secondVal = at.second.value;
```
madf: `matchValue`. And you need only one variable.

madf (changes requested):

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -121,32 +121,55 @@ void Server::handleReceive(const error_code& error, const std::optional<RadProto
...
 
-    User* user = nullptr;
-    if (m_users->FindByName(login, &user))
-    {
-        m_logger("User '%s' not found.", login.c_str());
-        printfd(__FILE__, "User '%s' NOT found!\n", login.c_str());
-        return nullptr;
+            auto requestAttrValue = attribute->toString();
+
+            std::string matchValue;
```
madf:
```c++
auto matchValue = at.second.value;
if (...)
    matchValue = std::to_string(m_dictionaries.attributeValueCode(requestAttrName, at.second.value));
```

madf (changes requested):

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -121,32 +121,53 @@ void Server::handleReceive(const error_code& error, const std::optional<RadProto
...
-    if (m_users->FindByName(login, &user))
-    {
-        m_logger("User '%s' not found.", login.c_str());
-        printfd(__FILE__, "User '%s' NOT found!\n", login.c_str());
-        return nullptr;
+            auto requestAttrValue = attribute->toString();
+
+            auto matchValue = at.second.value;
+
+            if (at.second.type != Config::AttrValue::Type::PARAM_NAME && m_dictionaries.attributeValueFindByName(requestAttrName, at.second.value))
```
madf:
```c++
if (at.second.type == Config::AttrValue::Type::VALUE ...
```

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -121,32 +121,53 @@ void Server::handleReceive(const error_code& error, const std::optional<RadProto
...
+                matchValue =  std::to_string(m_dictionaries.attributeValueCode(requestAttrName, at.second.value));
+
+            if (at.second.type == Config::AttrValue::Type::VALUE && matchValue != requestAttrValue)
```
madf: You check the type for `VALUE` twice. It would be logical to move this check under a single `if`:
```c++
if (at.second.type != Config::AttrValue::Type::PARAM_NAME)
{
    auto matchValue = at.second.value;
    if (<attribute-is-integer>)
        matchValue = <lookup-in-dictionary>;
    if (matchValue != requestAttrValue)
        return nullptr;
    break;
}
```

OTOH, your code looks like:
```c++
auto attrName = it.first;
for (const auto* attribute : packet.attributes())
{
    if (attrName != <attribute-name>)
        continue;
    if (<value>)
    {
        // return or break
    }
    // emplace back and break
}
```

So the inner cycle is just a search for the right attribute. And, besides, there is nothing outside the cycle. If the attribute is not found, it is skipped, which is wrong. If the attribute is not found, we should return `nullptr`. So instead, you should do something like this:
```c++
auto it = std::find_if(packet.attributes().begin(), packet.attributes().end(), <compare-by-name>);
if (it == packet.attributes().end())
    return nullptr; // Attribute is not found
auto* attribute = *it;
if (<is-value>)
{
    // if not match, return nullptr
}
else
{
    // emplace back
}
```

projects/stargazer/plugins/other/radius/server.cpp
madf: But see below...

madf (changes requested):

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -121,32 +122,55 @@ void Server::handleReceive(const error_code& error, const std::optional<RadProto
...
-    if (m_users->FindByName(login, &user))
-    {
-        m_logger("User '%s' not found.", login.c_str());
-        printfd(__FILE__, "User '%s' NOT found!\n", login.c_str());
-        return nullptr;
+        if (it == packet.attributes().end())
+            return nullptr;
+        auto* attribute = *it;
+
+        const std::string requestAttrName = m_dictionaries.attributeName(attribute->code());
```
madf: `const auto`?

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -121,32 +122,55 @@ void Server::handleReceive(const error_code& error, const std::optional<RadProto
...
+        auto requestAttrValue = attribute->toString();
```
madf: `const`?

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -121,32 +122,55 @@ void Server::handleReceive(const error_code& error, const std::optional<RadProto
...
+        auto matchValue = at.second.value;
+
+        if (at.second.type != Config::AttrValue::Type::PARAM_NAME)
+        {
+            if (m_dictionaries.attributeValueFindByName(requestAttrName, at.second.value))
+                matchValue =  std::to_string(m_dictionaries.attributeValueCode(requestAttrName, at.second.value));
```
madf: Double whitespace after `==`.

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -121,32 +122,55 @@ void Server::handleReceive(const error_code& error, const std::optional<RadProto
...
-        return nullptr;
+        if (it == packet.attributes().end())
+            return nullptr;
+        auto* attribute = *it;
+
+        const std::string requestAttrName = m_dictionaries.attributeName(attribute->code());
+        auto requestAttrValue = attribute->toString();
+        auto matchValue = at.second.value;
+
+        if (at.second.type != Config::AttrValue::Type::PARAM_NAME)
```
madf: Consider `at.second.type == Config::AttrValue::Type::VALUE`

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -121,32 +122,55 @@ void Server::handleReceive(const error_code& error, const std::optional<RadProto
...
+            if (m_dictionaries.attributeValueFindByName(requestAttrName, at.second.value))
+                matchValue =  std::to_string(m_dictionaries.attributeValueCode(requestAttrName, at.second.value));
+
+            if (matchValue != requestAttrValue)
+                return nullptr;
+        }
+        else
+        {
+            valuesForCompare.emplace_back(requestAttrValue, matchValue);
+            continue;
```
madf: Why `continue`?

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -121,32 +122,55 @@ void Server::handleReceive(const error_code& error, const std::optional<RadProto
...
-    if (password != user->GetProperties().password.Get())
+    while (m_users->SearchNext(h, &u) == 0)
     {
-        m_logger("User's password is incorrect.");
-        printfd(__FILE__, "User's password is incorrect.\n");
-        return nullptr;
+        bool allParamsMatch = true;
+        for (const auto& kv : valuesForCompare)
+        {
+            allParamsMatch = allParamsMatch && kv.first == u->GetParamValue(kv.second);
```
madf: I think, you need to pass `u->GetParamValue(kv.second)` through `m_dictionaries.attributeValueCode(...)` for integer attributes.

madf (changes requested):

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -121,32 +122,64 @@ void Server::handleReceive(const error_code& error, const std::optional<RadProto
...
-    std::string login;
-    std::string password;
-    for (const auto& attribute : packet.attributes())
+    std::vector<std::pair<std::pair<std::string, std::string>, std::string>> valuesForCompare;
+
+    for (const auto& at : m_config.getAuth().match)
     {
-        if (attribute->code() == RadProto::USER_NAME)
-            login = attribute->toString();
+        std::string matchName = at.first;
```
madf: Better to call it `attrName`.

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -121,32 +122,64 @@ void Server::handleReceive(const error_code& error, const std::optional<RadProto
...
-    if (m_users->FindByName(login, &user))
-    {
-        m_logger("User '%s' not found.", login.c_str());
-        printfd(__FILE__, "User '%s' NOT found!\n", login.c_str());
-        return nullptr;
+        if (it == packet.attributes().end())
+            return nullptr;
+        auto* attribute = *it;
+
+        const auto requestAttrName = m_dictionaries.attributeName(attribute->code());
```
madf: `requestAttrName` is the same as the `attrName` (now `matchName`) above.

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -121,32 +122,64 @@ void Server::handleReceive(const error_code& error, const std::optional<RadProto
...
+        const auto requestAttrValue = attribute->toString();
+        auto matchValue = at.second.value;
+
+        if (at.second.type == Config::AttrValue::Type::VALUE)
+        {
+            const auto matchType = m_dictionaries.attributeType(matchName);
+
+            if (matchType == "integer" && m_dictionaries.attributeValueFindByName(matchName, at.second.value))
```
madf: You can use `matchValue` instead of `at.second.value` here and below.

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -121,32 +122,64 @@ void Server::handleReceive(const error_code& error, const std::optional<RadProto
...
+                matchValue = std::to_string(m_dictionaries.attributeValueCode(matchName, at.second.value));
+
+            if (m_dictionaries.attributeValueFindByName(requestAttrName, at.second.value))
```
madf: What do these two lines do? This one and the one below?

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -121,32 +122,64 @@ void Server::handleReceive(const error_code& error, const std::optional<RadProto
...
+                matchValue = std::to_string(m_dictionaries.attributeValueCode(requestAttrName, at.second.value));
+
+            if (matchValue != requestAttrValue)
+                return nullptr;
+        }
+        else
+            valuesForCompare.emplace_back(std::make_pair(requestAttrName, requestAttrValue), matchValue);
```
madf: I don't think you need `requestAttrName` here. You better pack the attribute type (`matchType`) - you need it during the comparison later.

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -121,32 +122,64 @@ void Server::handleReceive(const error_code& error, const std::optional<RadProto
...
-        m_logger("User's password is incorrect.");
-        printfd(__FILE__, "User's password is incorrect.\n");
-        return nullptr;
+        bool allParamsMatch = true;
+        for (const auto& kv : valuesForCompare)
+        {
+            std::string paramValue = u->GetParamValue(kv.second);
+            const auto paramType = m_dictionaries.attributeType(kv.first.first);
+
+            if (paramType == "integer" && m_dictionaries.attributeValueFindByName(kv.first.first, kv.second))
```
madf: `kv.second` here is a user param name, no sense to look for it in the dictionary - it is not there.

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -121,32 +122,64 @@ void Server::handleReceive(const error_code& error, const std::optional<RadProto
...
+                paramValue = std::to_string(m_dictionaries.attributeValueCode(kv.first.first, kv.second));
```
madf: `kv.second` is a user param name, an you need to convert the param value, which is `paramValue`.

madf (changes requested):

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -121,32 +122,59 @@ void Server::handleReceive(const error_code& error, const std::optional<RadProto
...
-    if (m_users->FindByName(login, &user))
-    {
-        m_logger("User '%s' not found.", login.c_str());
-        printfd(__FILE__, "User '%s' NOT found!\n", login.c_str());
-        return nullptr;
+        if (it == packet.attributes().end())
+            return nullptr;
+        auto* attribute = *it;
+
+        attrName = m_dictionaries.attributeName(attribute->code());
```
madf: `attrName` already contains attribute name from the config. You've already looked it through a list of attributes by passing each attribute's code through the dictionary, any you've already found the `attribute` by its name, by matching it with `attrName`. You don't need to reassign it. Its like:
```c++
int a = 5;
a = 5;
```

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -121,32 +122,59 @@ void Server::handleReceive(const error_code& error, const std::optional<RadProto
...
+        const auto requestAttrValue = attribute->toString();
+        auto matchValue = at.second.value;
+        const auto matchType = m_dictionaries.attributeType(attrName);
```
madf: `attrType`. There are no types in `match`, only attributes have types.

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -121,32 +122,59 @@ void Server::handleReceive(const error_code& error, const std::optional<RadProto
...
     {
-        m_logger("User's password is incorrect.");
-        printfd(__FILE__, "User's password is incorrect.\n");
-        return nullptr;
+        bool allParamsMatch = true;
+        for (const auto& kv : valuesForCompare)
+        {
+            std::string paramValue = u->GetParamValue(kv.second);
+
+            if (kv.first.first == "integer" && m_dictionaries.attributeValueFindByName(kv.first.first, paramValue))
```
madf: `m_dictionaries.attributeValueFindByName(kv.first.first, paramValue)` - but `kv.first.first` is attribute type, not attribute name. I think, you need to provide attribute type, attribute name and attribute value here.

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -121,32 +122,59 @@ void Server::handleReceive(const error_code& error, const std::optional<RadProto
...
+                paramValue = std::to_string(m_dictionaries.attributeValueCode(kv.first.first, paramValue));
```
madf: Same as above, `kv.first.first` is attribute type, not name.

madf (changes requested):

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -122,7 +122,7 @@ void Server::handleReceive(const error_code& error, const std::optional<RadProto
 
 const User* Server::findUser(const RadProto::Packet& packet)
 {
-    std::vector<std::pair<std::pair<std::string, std::string>, std::string>> valuesForCompare;
+    std::vector<std::pair<std::pair<std::string,std::string>, std::pair<std::string, std::string>>> valuesForCompare;
```
madf: A vector... of pairs... of pairs... Isn't it look like too much pairs? Like, if in 2 years I would ask you: "What does it->second.first mean?". I am pretty sure you wouldn't be able to answer. And so do I.

Here you want to store information stat will be used when you iterate through users. You store attribute name, attribute type, attribute value and user param name. How about putting it all into a nice struct?

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -121,32 +122,58 @@ void Server::handleReceive(const error_code& error, const std::optional<RadProto
...
+
+            if (matchValue != requestAttrValue)
+                return nullptr;
+        }
+        else
+            valuesForCompare.emplace_back(std::make_pair(std::make_pair(attrType, attrName), std::make_pair(requestAttrValue, matchValue)));
     }
 
-    printfd(__FILE__, "User '%s' FOUND!\n", user->GetLogin().c_str());
+    User* u;
```
madf: `= nullptr;`

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -121,32 +122,58 @@ void Server::handleReceive(const error_code& error, const std::optional<RadProto
...
+    int h = m_users->OpenSearch();
```
madf: `const int` or `const auto`.

madf (changes requested):

projects/stargazer/plugins/other/radius/server.h
```diff
@@ -20,6 +20,15 @@ namespace STG
         public:
             Server(boost::asio::io_context& io_context, const std::string& secret, uint16_t port, const std::string& filePath, std::stop_token token, PluginLogger& logger, Users* users, const Config& config);
             void stop();
+
+            struct DataVector
```
madf: No need to have it in the header, you use it only in a single method, inside its implementation. Put it in the .cpp in the anonymous namespace or even inside the function.

madf: Also, let's call it `DataForCompare`

madf (changes requested):

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -121,32 +122,66 @@ void Server::handleReceive(const error_code& error, const std::optional<RadProto
...
-    User* user = nullptr;
-    if (m_users->FindByName(login, &user))
+    for (const auto& at : m_config.getAuth().match)
     {
-        m_logger("User '%s' not found.", login.c_str());
-        printfd(__FILE__, "User '%s' NOT found!\n", login.c_str());
-        return nullptr;
+        std::string attrName = at.first;
+
+        auto it = std::find_if(packet.attributes().begin(), packet.attributes().end(), [this, attrName](const auto& atr){return attrName == m_dictionaries.attributeName(atr->code());});
```
madf: Here you have two cycles: one by attribetes from the request (inside the `std::find_if`), and another one by dictionary. You do name lookup for each packet attribute to compare it with the name from the `match`. I think, you can do the dictionary lookup only once: you can find a code for the attribute name from `match` and then do simple code comparison in the `std::find_if`. Besides, when you compare two strings (in your case - attribute names) - it is also done by a cycle. But when you compare codes, they are just numbers! This is a single CPU instruction.

madf (changes requested):

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -121,32 +122,67 @@ void Server::handleReceive(const error_code& error, const std::optional<RadProto
...
-        return nullptr;
+        uint32_t attrCode = m_dictionaries.attributeCode(at.first);
+
+        auto it = std::find_if(packet.attributes().begin(), packet.attributes().end(), [this, attrCode](const auto& atr){return attrCode == atr->code();});
+
+        if (it == packet.attributes().end())
+            return nullptr;
+        auto* attribute = *it;
+
+        std::string attrName = at.first;
```
madf: You can use it in `m_dictionaries.attributeCode(at.first);` instead of `at.first`, if you declare it earlier. Also, since it does not change, you can mar it as `const`.

madf (changes requested):

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -121,32 +122,67 @@ void Server::handleReceive(const error_code& error, const std::optional<RadProto
...
 
-    User* user = nullptr;
-    if (m_users->FindByName(login, &user))
+    for (const auto& at : m_config.getAuth().match)
     {
-        m_logger("User '%s' not found.", login.c_str());
-        printfd(__FILE__, "User '%s' NOT found!\n", login.c_str());
-        return nullptr;
+        const std::string attrName = at.first;
+        uint32_t attrCode = m_dictionaries.attributeCode(attrName);
```
madf: Please add `const`.

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -121,32 +122,67 @@ void Server::handleReceive(const error_code& error, const std::optional<RadProto
...
+
+        auto it = std::find_if(packet.attributes().begin(), packet.attributes().end(), [this, attrCode](const auto& atr){return attrCode == atr->code();});
```
madf: Please add `const`.

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -121,32 +122,67 @@ void Server::handleReceive(const error_code& error, const std::optional<RadProto
...
+
+        if (it == packet.attributes().end())
+            return nullptr;
+        auto* attribute = *it;
```
madf: Please add `const`.

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -121,32 +122,67 @@ void Server::handleReceive(const error_code& error, const std::optional<RadProto
...
+
+        const auto requestAttrValue = attribute->toString();
+        auto matchValue = at.second.value;
+        const auto attrType = m_dictionaries.attributeType(attrName);
```
madf: You can lookup attribute type by code, rather than by name, which is slightly more efficient.

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -121,32 +122,67 @@ void Server::handleReceive(const error_code& error, const std::optional<RadProto
...
-    std::string password;
-    for (const auto& attribute : packet.attributes())
+    struct DataForCompare
     {
-        if (attribute->code() == RadProto::USER_NAME)
-            login = attribute->toString();
+        std::string attrType;
+        std::string attrName;
+        std::string requestAttrValue;
+        std::string matchValue;
```
madf: Let's call it `paramName`.

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -121,32 +122,67 @@ void Server::handleReceive(const error_code& error, const std::optional<RadProto
...
+
+        const auto it = std::find_if(packet.attributes().begin(), packet.attributes().end(), [this, attrCode](const auto& atr){return attrCode == atr->code();});
+
+        if (it == packet.attributes().end())
+            return nullptr;
+        const auto* attribute = *it;
+
+        const auto requestAttrValue = attribute->toString();
+        auto paramName = at.second.value;
+        const auto attrType = m_dictionaries.attributeType(attrCode);
```
madf: You look for `attrCode` and `attrType` for the attribute name from the `match` section on every incoming request. This is suboptimal, the attributes does not change often. It would be better to do it while you parse the config, only once. It requires changing `Config::ASection::Pairs` to store attribute code and type along with the name. Likely you would need to change the `std::pair` to a `struct`.

projects/stargazer/plugins/other/radius/config.h
```diff
@@ -27,30 +28,39 @@ namespace STG
                 Type type;
             };
 
+            struct ASectionData
```
madf: You may move `ASectionData` as a `Data` subtype in `ASection`.

projects/stargazer/plugins/other/radius/config.h
```diff
@@ -27,30 +28,39 @@ namespace STG
...
+                std::string attrType;
+                AttrValue value;
+            };
+
             struct ASection
             {
-                using Pairs = std::vector<std::pair<std::string, AttrValue>>;
-                Pairs match;
-                Pairs send;
+                using ASectionDataVect = std::vector<ASectionData>;
```
madf: When you move `ASectionData` into `ASection::Data` you may call this type `Datas`.

projects/stargazer/plugins/other/radius/config.h
```diff
@@ -27,30 +28,39 @@ namespace STG
...
+                ASectionDataVect match;
+                ASectionDataVect send;
             };
 
             const std::string& GetStrError() const { return m_errorStr; }
             int ParseSettings(const ModuleSettings& s);
 
             uint16_t GetPort() const { return m_port; }
-            const std::string& GetDictionaries() const { return m_dictionaries; }
+            const std::string& GetDictionaries() const { return m_dictionariesPath; }
```
madf: Let's call the function `GetDictionariesPath`.

projects/stargazer/plugins/other/radius/server.cpp
```diff
@@ -52,12 +52,12 @@ std::vector<RadProto::Attribute*> Server::makeAttributes(const User* user)
...
-            attrValue = user->GetParamValue(at.second.value);
+        if (at.value.type == Config::AttrValue::Type::PARAM_NAME)
+            attrValue = user->GetParamValue(at.value.value);
         else
-            attrValue = at.second.value;
+            attrValue = at.value.value;
 
-        const auto attrName = at.first;
+        const auto attrName = at.attrName;
         const auto attrCode = m_dictionaries.attributeCode(attrName);
```
madf: Aren't `attrCode` and `attrType` already in the `at`?

projects/stargazer/plugins/other/radius/server.cpp
madf: It seems that you no longer need the whole `m_dictionaries` object here, you have one in the `m_config`. You can expose it through a method and use it directly, or create a handy reference here.

madf (changes requested):

projects/stargazer/plugins/other/radius/config.h
```diff
@@ -28,32 +28,33 @@ namespace STG
...
-                ASectionDataVect send;
+               struct Data
+                {
+                    std::string attrName;
+                    uint32_t attrCode;
+                    std::string attrType;
+                    AttrValue value;
+                };
+
+                Data data;
```
madf: What does this line do?

madf:
Stargazer crashes on incorrect attribute name in the config.
The config:
```
    <auth>
        send = "foo = \'Login-User\'"
        match = "User-Password = password"
    </auth>
```
The crash:
```
/tmp/stg/projects/stargazer/users_impl.cpp > 16:06:08 > =====================| pid: 2796902 |===================== 
/tmp/stg/projects/stargazer/traffcounter_impl.cpp > 16:06:08 > TraffCounter::Start()
/tmp/stg/projects/stargazer/users_impl.cpp > 16:06:08 > Day = 3 Min = 6
/tmp/stg/projects/stargazer/users_impl.cpp > 16:06:08 > Monitor=0 file USERS /var/stargazer/monitoring/users_r
/tmp/stg/projects/stargazer/plugin_mgr.cpp > 16:06:08 > Module: projects/stargazer/plugins/mod_auth_ao.so
/tmp/stg/projects/stargazer/plugin_mgr.cpp > 16:06:08 > Module: projects/stargazer/plugins/mod_cap_ether.so
/tmp/stg/projects/stargazer/plugin_mgr.cpp > 16:06:08 > Module: projects/stargazer/plugins/mod_auth_ia.so
/tmp/stg/projects/stargazer/plugin_runner.cpp > 16:06:08 > PluginRunner::load() - Error loading plugin 'projects/stargazer/plugins/mod_auth_ia.so': 'projects/stargazer/plugins/mod_auth_ia.so: undefined symbol: stgTime'
/tmp/stg/projects/stargazer/plugin_mgr.cpp > 16:06:08 > Error loading plugin 'projects/stargazer/plugins/mod_auth_ia.so': 'projects/stargazer/plugins/mod_auth_ia.so: undefined symbol: stgTime'
/tmp/stg/projects/stargazer/plugin_mgr.cpp > 16:06:08 > Module: projects/stargazer/plugins/mod_ping.so
/tmp/stg/projects/stargazer/plugin_mgr.cpp > 16:06:08 > Module: projects/stargazer/plugins/mod_radius.so
/tmp/stg/projects/stargazer/traffcounter_impl.cpp > 16:06:08 > Monitor=1 file TraffCounter /var/stargazer/monitoring/traffcounter_r
terminate called after throwing an instance of 'std::out_of_range'
  what():  map::at
Aborted                    sudo ./projects/stargazer/stargazer -f ../dist/
```

madf:
Stargazer crashes on incorrect attribute value.
The config:
```
    <auth>
        send = "Service-Type = \'"
        match = "User-Password = foo"
    </auth>
```
The crash:
```
/tmp/stg/projects/stargazer/users_impl.cpp > 16:10:35 > =====================| pid: 2797917 |===================== 
/tmp/stg/projects/stargazer/users_impl.cpp > 16:10:35 > Day = 3 Min = 10
/tmp/stg/projects/stargazer/traffcounter_impl.cpp > 16:10:35 > TraffCounter::Start()
/tmp/stg/projects/stargazer/users_impl.cpp > 16:10:35 > Monitor=0 file USERS /var/stargazer/monitoring/users_r
/tmp/stg/projects/stargazer/plugin_mgr.cpp > 16:10:35 > Module: projects/stargazer/plugins/mod_auth_ao.so
/tmp/stg/projects/stargazer/plugin_mgr.cpp > 16:10:35 > Module: projects/stargazer/plugins/mod_cap_ether.so
/tmp/stg/projects/stargazer/plugin_mgr.cpp > 16:10:35 > Module: projects/stargazer/plugins/mod_auth_ia.so
/tmp/stg/projects/stargazer/plugin_runner.cpp > 16:10:35 > PluginRunner::load() - Error loading plugin 'projects/stargazer/plugins/mod_auth_ia.so': 'projects/stargazer/plugins/mod_auth_ia.so: undefined symbol: stgTime'
/tmp/stg/projects/stargazer/plugin_mgr.cpp > 16:10:35 > Error loading plugin 'projects/stargazer/plugins/mod_auth_ia.so': 'projects/stargazer/plugins/mod_auth_ia.so: undefined symbol: stgTime'
/tmp/stg/projects/stargazer/plugin_mgr.cpp > 16:10:35 > Module: projects/stargazer/plugins/mod_ping.so
/tmp/stg/projects/stargazer/plugin_mgr.cpp > 16:10:35 > Module: projects/stargazer/plugins/mod_radius.so
/tmp/stg/projects/stargazer/traffcounter_impl.cpp > 16:10:35 > Monitor=1 file TraffCounter /var/stargazer/monitoring/traffcounter_r
terminate called after throwing an instance of 'std::out_of_range'
  what():  basic_string::erase: __pos (which is 18446744073709551615) > this->size() (which is 0)
Aborted                    sudo ./projects/stargazer/stargazer -f ../dist/
```

madf:
Stargazer crashes on missing config value:
The config:
```
    <auth>
        send =
        match = "User-Password = foo"
    </auth>
```
The crash:
```
/tmp/stg/projects/stargazer/users_impl.cpp > 16:12:25 > =====================| pid: 2798281 |===================== 
/tmp/stg/projects/stargazer/users_impl.cpp > 16:12:25 > Day = 3 Min = 12
/tmp/stg/projects/stargazer/users_impl.cpp > 16:12:25 > Monitor=0 file USERS /var/stargazer/monitoring/users_r
/tmp/stg/projects/stargazer/traffcounter_impl.cpp > 16:12:25 > TraffCounter::Start()
/tmp/stg/projects/stargazer/plugin_mgr.cpp > 16:12:25 > Module: projects/stargazer/plugins/mod_auth_ao.so
/tmp/stg/projects/stargazer/plugin_mgr.cpp > 16:12:25 > Module: projects/stargazer/plugins/mod_cap_ether.so
/tmp/stg/projects/stargazer/plugin_mgr.cpp > 16:12:25 > Module: projects/stargazer/plugins/mod_auth_ia.so
/tmp/stg/projects/stargazer/plugin_runner.cpp > 16:12:25 > PluginRunner::load() - Error loading plugin 'projects/stargazer/plugins/mod_auth_ia.so': 'projects/stargazer/plugins/mod_auth_ia.so: undefined symbol: stgTime'
/tmp/stg/projects/stargazer/plugin_mgr.cpp > 16:12:25 > Error loading plugin 'projects/stargazer/plugins/mod_auth_ia.so': 'projects/stargazer/plugins/mod_auth_ia.so: undefined symbol: stgTime'
/tmp/stg/projects/stargazer/plugin_mgr.cpp > 16:12:25 > Module: projects/stargazer/plugins/mod_ping.so
/tmp/stg/projects/stargazer/plugin_mgr.cpp > 16:12:25 > Module: projects/stargazer/plugins/mod_radius.so
/tmp/stg/projects/stargazer/traffcounter_impl.cpp > 16:12:26 > Monitor=1 file TraffCounter /var/stargazer/monitoring/traffcounter_r
/usr/lib/gcc/x86_64-pc-linux-gnu/15/include/g++-v15/bits/stl_vector.h:1282: std::vector<_Tp, _Alloc>::const_reference std::vector<_Tp, _Alloc>::operator[](size_type) const [with _Tp = std::__cxx11::basic_string<char>; _Alloc = std::allocator<std::__cxx11::basic_string<char> >; const_reference = const std::__cxx11::basic_string<char>&; size_type = long unsigned int]: Assertion '__n < this->size()' failed.
Aborted                    sudo ./projects/stargazer/stargazer -f ../dist/
```

madf:
Stargazer ignores the path to dictionaries.
The config:
```
    Dictionaries = ./dictionary
```
There is no such file, but Stargazer starts successfully and serves RADIUS requests. I guess, it just uses the hard-coded default path.

projects/stargazer/plugins/other/radius/config.cpp
```diff
@@ -72,7 +72,13 @@ std::vector<Data> Config::ParseRules(const std::string& value, const std::string
         }
 
         const std::string attrName = keyValue[0];
-        const uint32_t attrCode = m_dictionaries.attributeCode(attrName);
+        uint32_t attrCode;
+
+        if (m_dictionaries.attributeFindByName(attrName))
+            attrCode = m_dictionaries.attributeCode(attrName);
+        else
+            return {};
```
madf: It's better to tell here which exact attribute name caused the problem, rather than silently return empty rules.

projects/stargazer/plugins/other/radius/config.cpp
```diff
@@ -43,60 +44,74 @@ std::vector<Data> Config::ParseRules(const std::string& value, const std::string
...
-        const auto attrType = m_dictionaries.attributeType(attrCode);
+            const auto attrType = m_dictionaries.attributeType(attrCode);
 
-        res.push_back({attrName, attrCode, attrType, {valueName, type}});
+            res.push_back({attrName, attrCode, attrType, {valueName, type}});
+        }
+    }
+    catch (const std::exception& e)
+    {
+        printfd(__FILE__, "Error ParseRules: incorrect parameter value. Exception: '%s'\n", e.what());
```
madf: Here we don't actually know what exactly caused the exception. So the wording should be something like this: `printfd(__FILE__, "Error in ParseRules. %s", e.what());`

projects/stargazer/plugins/other/radius/config.cpp
```diff
@@ -43,60 +44,74 @@ std::vector<Data> Config::ParseRules(const std::string& value, const std::string
...
+    {
+        const auto mit = std::find(conf.begin(), conf.end(), ParamValue("match", {}));
+        if (mit != conf.end())
+            res.match = ParseRules(mit->value.at(0), mit->param);
 
+        const auto sit = std::find(conf.begin(), conf.end(), ParamValue("send", {}));
+        if (sit != conf.end())
+            res.send = ParseRules(sit->value.at(0), sit->param);
+    }
+    catch(const std::out_of_range& e)
```
madf: The `ParseRules` catches everything by `std::exception`, so `std::out_of_range` can not happen here. The whole try/catch block here is useless.

projects/stargazer/plugins/other/radius/config.cpp
```diff
@@ -154,7 +154,7 @@ int Config::ParseSettings(const ModuleSettings & s)
     pv.param = "Dictionaries";
     pvi = std::find(s.moduleParams.begin(), s.moduleParams.end(), pv);
     if (pvi != s.moduleParams.end() && !pvi->value.empty())
-        m_dictionariesPath = pvi->value[0];
+        m_dictionaries = pvi->value[0];
```
madf: But... `pvi->value[0]` is a string, and `m_dictionaries` is an object of the `RadProto::Dictionaries` class... How can you simply assign?

madf (changes requested):

